PR #441, opened 2020-08-03
No inverse relationship crashes transport layer

### LINKED ISSUE #437: No inverse relationship crashes transport layer
If you have an ontology with a relationship that has no inverse defined, the instantiation fails on server side of the transport session. An example to reproduce: 

- Take the city ontology and out comment the inverse of `hasInhabitant` and `isInhabitantOf` (i.e. `hasInhabitant` has no inverse from that point on) 
- Reinstall the city ontology
- Run the unit test `python -m unittest tests/test_transport_sqlite_city.py`

Error:
```
RuntimeError("Error on Server side: %s" % data[7:])
RuntimeError: Error on Server side: KeyError: "Unknown entity 'INVERSE_OF_hasInhabitant' in namespace city. For backwards compatibility reasons we also looked for I_N_V_E_R_S_E__O_F_HAS_INHABITANT and failed."
```

urbanmatthias:
Fixes #437

### DIFF
--- a/osp/core/ontology/namespace.py
+++ b/osp/core/ontology/namespace.py
@@ -126,6 +126,22 @@ class OntologyNamespace():
         :return: The ontology entity
         :rtype: OntologyEntity
         """
+        try:
+            return self._do_get(name, _case_sensitive, _force_by_iri)
+        except KeyError as e:
+            if name.startswith("INVERSE_OF_"):
+                return self._do_get(name[11:], _case_sensitive,
+                                    _force_by_iri).inverse
+            raise e
+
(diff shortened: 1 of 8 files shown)


osp/core/ontology/namespace.py
```diff
@@ -121,6 +121,22 @@ def get(self, name, fallback=None):
     def _get(self, name, _case_sensitive=False, _force_by_iri=False):
         """Get an ontology entity from the registry by name.
 
+        :param name: The name of the ontology entity
```
urbanmatthias, line 124: @aaronAB1993

PR #465, opened 2020-08-10
Delete cuds ubject recursively

urbanmatthias:


osp/core/session/session.py
```diff
@@ -76,7 +76,7 @@ def delete_cuds_object(self, cuds_object):
         from osp.core.namespaces import cuba
         if cuds_object.session != self:
             cuds_object = next(self.load(cuds_object.uid))
-        if cuds_object.get():
+        if cuds_object.get(rel=cuba.relationship):
```
urbanmatthias, line 79: `get()` without argument returns less than `cuds_object.get(rel=cuba.relationship)`, because default value of rel is `cuba.activeRelationship`

Therefore sometimes it didn't remove passive relationships.

In a setting where the objects related by passive relationships are deleted afterwards, an error occurs, because it tries to modify the object that has been deleted before, by trying to remove the relationship that should have been deleted before.

PR #468, opened 2020-08-11
Add easy way to serialize CUDS objects

### LINKED ISSUE #322: Add easy way to serialize CUDS objects
In GitLab by @urbanmatthias on Jun 8, 2020, 17:42

- update `__getstate__`
- implemet a `to_json` method
- add a method to utils

urbanmatthias:
Fixes #322

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -4,6 +4,7 @@ import responses
 import uuid
 import os
 import osp.core
+import json
 from osp.core.namespaces import cuba
 from osp.core.session.transport.transport_utils import serializable
 from osp.core.session.core_session import CoreSession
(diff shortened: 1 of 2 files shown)


yoavnash:
How is it now easier?

urbanmatthias:
There are deserialize and serialize functions in osp.core.utils, which can be called with a cuds objects only.

No need to specify a session, or call find_cuds_object. Although it is still a allowed to specify a session, if the user wants to.

tests/test_utils.py
```diff
@@ -201,6 +217,31 @@ def test_deserialize(self):
...
+
+    def test_serialize(self):
+        c = branch(
+            city.City(name="Freiburg", uid=1),
+            branch(
+                city.Neighborhood(name="Littenweiler", uid=2),
+                city.Street(name="Schwarzwaldstraße", uid=3)
+            )
+        )
+        self.maxDiff = None
```
yoavnash, line 236: What does this do?

urbanmatthias: The self.maxDiff? In case there is an error in the unittest, it will print the entire diff and not only one line

yoavnash:
Alright, great. Then I think this together with the openAPI document should be enough for now, right?

urbanmatthias:
I hope so

PR #474, opened 2020-08-11
V3.4.1 Patch

urbanmatthias:
There happened some important fixes and additions that should be released:
- Added the `yaml2camelcase` to convert the entity names of yaml ontologies to camel case.
- Fixed ontology2dot
- Updated an example with postgresql
- Fixed a bug when trying to clear an empty database
- Improved error messages of pico
- Add easy way to serialize CUDS objects: osp.core.utils.serialize / desierialize
- added method to delete cuds objects recursively and fixed a bug that made this impossible

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
           radon mi -s .
         
       - name: bandit
-        run: bandit -r osp --skip B101
+        run: bandit -r osp --skip B101,B322
   
   test:
     runs-on: self-hosted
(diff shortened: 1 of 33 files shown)


yoavnash:
Shouldn't we increase the version number to 3.5.1?

urbanmatthias:
No, because the last release was v3.4.0.

Maybe the naming of the branch is bad. It was v3.4.0-dev when we worked on the v.3.4.0-dev milestone, where all intermediate releases where called 3.3.X-beta.

No we work on the (not yet existing) milestone v3.5.0, so the branch is called v3.5.0-dev.

urbanmatthias:
I renamed the Pull Request

urbanmatthias:
- Cleaned up source code files
- Fixed utils.getrdfgraph
- default relationship can now refer to different namespace

PR #477, opened 2020-08-11
fixed get rdfgrpah

aaronAB1993:


osp/core/utils/general.py
```diff
@@ -55,13 +56,19 @@ def get_rdf_graph(session=None):
...
+            )
     from osp.core.cuds import Cuds
-    from osp.core import ONTOLOGY_NAMESPACE_REGISTRY
+    from osp.core.namespaces import _namespace_registry
     session = session or Cuds._session
     graph = rdflib.Graph()
     for triple in session.get_triples():
         graph.add(triple)
-    for namespace in ONTOLOGY_NAMESPACE_REGISTRY:
+    for namespace in _namespace_registry:
```
urbanmatthias, line 71: Since we already have the graph of the namespace_registry, this looks unnecessarily inefficient.
I guess we should fiy this as well.

What do you think @aaronAB1993

urbanmatthias: I mean that we iterate over all entities and call get_triples

aaronAB1993: So what you suggest is to merge the namespace registry graph with the "CUDS graph" and then return?

urbanmatthias: Exactly, although I'm not sure how to `merge two graphs`  ideally.
I've read in the rdflib docs, that using the python union operator can cause problems.

aaronAB1993: Yes but only for blank nodes, right? Do we expect any blank nodes for the CUDS graph?

urbanmatthias: Hmm you are right. There shouldn't be any in the CUDS graph.

urbanmatthias: Then this should actually be pretty straighforward, right?

aaronAB1993: I fixed this in the latest commit. It is better now.

Issue #227, opened 2020-06-22
Set a relationship from a different namespace as default


pablo-de-andres:
In GitLab by @yoavnash on Jan 7, 2020, 16:28

Currently it's only possible to set a relationship from the same namespace as the default one.

aaronAB1993:
@yoavnash you can now refer to a relationship from a different namespace:

```yml
---
    version: "0.0.3"
    
    namespace: "default_rel_test_namespace_two_definitions"
    default_rel: cuba.activeRelationship
    ontology:
      SomeEntity:
        subclass_of:
          - cuba.Entity
      someRel:
        subclass_of:
          - cuba.activeRelationship
```

The "old" way to mark a default relationship is still valid:
```yml
---
    version: "0.0.3"
    
    namespace: "default_rel_test_namespace_two_definitions"
    ontology:
      SomeEntity:
        subclass_of:
          - cuba.Entity
      someRel:
        subclass_of:
          - cuba.activeRelationship
        default_rel: true
```
also works. However, when you do both, it raises an error, because defining the default twice doesn't make any sense.

yoavnash:
Great!

PR #478, opened 2020-08-11
default relationship can now refer to different namespace

aaronAB1993:
#227

### DIFF
--- a/osp/core/ontology/yml/yml_parser.py
+++ b/osp/core/ontology/yml/yml_parser.py
@@ -87,7 +87,7 @@ class YmlParser:
             t = self._validate_entity(entity_name, entity_doc)
             types[entity_name] = t
 
-        self._assert_default_relationship_occurance()
+        self._assert_default_relationship_occurrence()
         self._check_default_rel_definition_on_ontology()
         for entity_name, entity_doc in self._ontology_doc.items():
             # self._load_class_expressions(entity) TODO
@@ -406,21 +406,22 @@ class YmlParser:
             (iri, rdflib.OWL.inverseOf, inverse_iri)
         )
 
-    def _assert_default_relationship_occurance(self):
+    def _assert_default_relationship_occurrence(self):
         """Assures that only one default relationship is defined in the yaml
 
         :raises ValueError: If more than one definition is found.
         """
-        occurances = 0
+        occurrences = 0
         if DEFAULT_REL_KEY in self._doc:
-            occurances += 1
+            occurrences += 1
         for entity_name, entity_doc in self._ontology_doc.items():
             if DEFAULT_REL_KEY in entity_doc:
-                occurances += 1
-        if occurances > 1:
-            raise ValueError(f"""You have defined {occurances} default
-        relationships for namespace {self._namespace}
-        although <= 1 are allowed.""")
+                occurrences += 1
+        if occurrences > 1:
+            raise ValueError(
+                f"You have defined {occurrences} default relationships for "
+                f"namespace {self._namespace} although <= 1 are allowed."
+            )
 
     def _check_default_rel_definition_on_ontology(self):
         """Check if the given yaml defines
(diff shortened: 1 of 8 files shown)


osp/core/ontology/yml/yml_parser.py
```diff
@@ -404,8 +406,51 @@ def _set_inverse(self, entity_name, entity_doc):
...
+        a default relationship, save that accordingly.
+        """
+        if DEFAULT_REL_KEY in self._doc:
+            namespace, entity_name = self._doc[DEFAULT_REL_KEY].split('.')
+
+            # defined relationship must be installed
+            from osp.core.namespaces import _namespace_registry
+            referred_namespace = _namespace_registry.get(namespace)
+            if not referred_namespace:
+                raise ValueError(f"""The namespace {namespace} that you
```
urbanmatthias, line 436: I prefer using
```py
ValueError(f"line1 "
           f"line2 "
           f"line3 ")
```
over 
```py
ValueError(f"""line1
    line2
    line3""")
```
if the string does not fit in one line.
The first string results in `line1 line2 line3` while the latter results in

```
line1
    line2
    line3
```

urbanmatthias: ValueError: You have defined 2 default relationships for namespace default_rel_test_namespace_two_definitions although <= 1 are allowed.

vs.

ValueError: You have defined 2 default
        relationships for namespace default_rel_test_namespace_two_definitions
        although <= 1 are allowed.

aaronAB1993: Alright, thanks :)

osp/core/ontology/yml/yml_parser.py
```diff
@@ -404,8 +406,51 @@ def _set_inverse(self, entity_name, entity_doc):
             (iri, rdflib.OWL.inverseOf, inverse_iri)
         )
 
-    def _check_default_rel(self, entity_name, entity_doc):
-        """Check if the given relationship the default
+    def _assert_default_relationship_occurance(self):
```
urbanmatthias, line 409: Occurrence is the correct way to spell this word, meaning 'an event that happened.'

Occurance is one of the most popular ways that this world is misspelt.

https://hinative.com/en-US/questions/1554753

Is this true?

urbanmatthias: I will update it to occurrence

aaronAB1993: True: https://www.dict.cc/?s=occurrence

But I am happy to be one out of millions :)

tests/test_namespace.py
```diff
@@ -312,10 +313,37 @@ def test_namespace_str(self):
                          "<city: http://www.osp-core.com/city#>")
 
     def test_get_default_rel(self):
+        # default rel defined as flag in entity name
         self.installer.install("city")
         namespace = self.namespace_registry.city
         self.assertEqual(namespace.get_default_rel().name, "hasPart")
 
+        onto_def_rel = os.path.join(
```
urbanmatthias, line 321: You're technically not testing the namespace or the get_default method, as the method name or the file name of this test might suggest.

But I guess that's fine. What do you think?

aaronAB1993: I think you are right, there is no need to install the namespace, i can rather directly test the `YmlParser.parse()`. I will put the test somewhere else.

aaronAB1993: Actually, I just found that it is ok. What I want to test is that the installation of the yml will result in the right default relationship in the namespace. I think the confusion is that the `YmlParser` is used for installation and this is where I put my code. So the question is, is this test part of the installation, the namespace or the parser? I could change the name of the test though.

urbanmatthias: Let it as it is. I think it is fine

urbanmatthias:
I think you can merge now

Issue #319, opened 2020-06-22
Make pico uninstall "*" more user friendly


pablo-de-andres:
In GitLab by @pablo-de-andres on Jun 5, 2020, 14:37

- Make the error for `pico uninstall *` (* is unquoted) more understandable.
- Maybe provide a keyword to used instead of "*", like `all`.

urbanmatthias:
* is impossible, because bash will automaically replace it with all files in the current directory.
I like `all` though.

aaronAB1993:
It is now `pico uninstall all` to uninstall all ontologies.

pablo-de-andres:
And `"*"` is no longer supported?
Asking for the docs

PR #490, opened 2020-08-14
introduces keyword all. made error message a bit clearer

aaronAB1993:
#319


aaronAB1993:
I did not change much about the error message for `pico uninstall *`. The reason is that it is most generic as it is right now.

aaronAB1993:
Where do I have to adapt the documentation for `pico`? It is now `pico uninstall all` instead of `pico uninstall "*"`.

PR #509, opened 2020-08-28
Change the style of the docstrings Fixes #253

### LINKED ISSUE #253: Change the style of the docstrings
In GitLab by @pablo-de-andres on Feb 12, 2020, 15:11

Currently, we follow an rst syntax in the docstrings:

```python
def add(x,y):
    """
    Adds two numbers.

    :param x: The first value to add
    :type x: int
    :param y: The second value to add
    :type y: int
    :returns: Result of the addition
    :rtype: int
    """
```

But it is not the most reader-friendly format.
There are alternatives, like the [Google Python Style Guide](https://google.github.io/styleguide/pyguide.html) which seem more readable:

```python
def add(x,y):
    """
    Adds two numbers.

    Args:
        x (int): The first value to add
        y (int): The second value to add
    Returns
        int : Result of the addition
    """
```

Google's Python Style Guide is also supported in Sphinx via the [napoleon extension](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html), ant there are extensions for IDEs that also support this format.

urbanmatthias:


urbanmatthias:
@pablo-de-andres I request review and compliments (Bitte schnell).

pablo-de-andres (approved):
Looks good! (Also, nice turtleneck)

PR #519, opened 2020-09-09
Hotfix: Issues with EMMO Fixes #518

urbanmatthias:

Issue #518, opened 2020-09-09
Issues with EMMO


urbanmatthias:
adding throws errors because inverse can be blank node.
some attributes are recognized as must-have but aren't

PR #520, opened 2020-09-10
Issues with EMMO

urbanmatthias:
There were several issues with EMMO: https://github.com/simphony/osp-core/issues/518

I already merged the hotfix to master: https://github.com/simphony/osp-core/pull/519


pablo-de-andres:
Does this fix the issues?

PR #524, opened 2020-09-25
Make SQLWrappers more similar to triplestores

### LINKED ISSUE #483: Make SQLWrappers more similar to triplestores
We want to support more and more standard semantic web tools like reasoning and SPARQL.
It may make sense to transform the table layout of the SQLWrappers to be more like Triplestores.

This could improve the import / export to rdf data needed for #297

urbanmatthias:
Fixes #483


yoavnash:
Perhaps to review it, it would make sense to set up a meeting and go over it together

yoavnash (approved):

PR #527, opened 2020-10-02
Allow to set install dir via env variable

urbanmatthias:
All the user to specify the install directory of ontologies in an environment variable: $OSP_ONTOLOGIES_DIR

### DIFF
--- a/osp/core/__init__.py
+++ b/osp/core/__init__.py
@@ -1,9 +1,16 @@
 import sys
+import rdflib
 import logging
-import osp.core.namespaces
-from osp.core.namespaces import _namespace_registry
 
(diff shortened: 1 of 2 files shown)


osp/core/__init__.py
```diff
@@ -1,10 +1,17 @@
 import sys
+import rdflib
 import logging
-import osp.core.namespaces
-from osp.core.namespaces import _namespace_registry
 
 logging.getLogger("rdflib").setLevel(logging.WARNING)
 
+try:
```
urbanmatthias, line 7: this is to make osp-core rdflib 4 compatible

pablo-de-andres (approved):
Looks good!

PR #528, opened 2020-10-02
Allow to set install dir via env variable

urbanmatthias:
All the user to specify the install directory of ontologies in an environment variable: $OSP_ONTOLOGIES_DIR


pablo-de-andres (approved):
Looks good!

PR #573, opened 2021-02-01
Query ontology restrictions

urbanmatthias:


kysrpex:
Just committed some tests for the Restriction class. The tests for the Composition class are still pending.

tests/test_restrictions_emmo.py
```diff
@@ -15,10 +23,137 @@
...
+        Also tests _compute_target and _check_quantifier implicitly.
+        """
+        for r in self.iter_restrictions():
+            b_node = r._bnode
+            graph = r._graph
+
+            restriction_predicates = graph.predicates(subject=b_node)
+            quantifiers = [self.quantifiers[p] for p in restriction_predicates
+                           if p in self.quantifiers]
+            self.assertTrue(len(quantifiers) == 1, "The restriction has not "
```
urbanmatthias, line 64: Hi @kysrpex 
in unittest2, you can use 
self.assertEqual(arg1, arg2) to check if arg1 equals arg2.
I think the benefit of using assertEqual compared to assertTrue is that the former prints more info when the two are not the same

tests/test_restrictions_emmo.py
```diff
@@ -0,0 +1,160 @@
...
+        for datatype in self.emmo_datatypes:
+            restrictions = itertools.chain(restrictions,
+                                           (r for r in datatype.axioms
+                                            if type(r) is Restriction))
+        return restrictions
+
+    def test___str__(self):
+        """Test the __str__ method."""
+        for r in self.iter_restrictions():
+            self.assertIs(type(r.__str__()), str)
```
urbanmatthias, line 50: It probably makes sense to also test the specific string value for some specific restrictions

tests/test_restrictions_emmo.py
```diff
@@ -0,0 +1,160 @@
...
+                   rdflib.OWL.minCardinality: QUANTIFIER.MIN,
+                   rdflib.OWL.maxCardinality: QUANTIFIER.MAX}
+
+    rtypes = {OntologyRelationship: RTYPE.RELATIONSHIP_RESTRICTION,
+              OntologyAttribute: RTYPE.ATTRIBUTE_RESTRICTION}
+
+    def iter_restrictions(self):
+        """Returns all the datatypes of the test wrapped in an iterator."""
+        restrictions = iter([])
+        for datatype in self.emmo_datatypes:
```
urbanmatthias, line 41: Restrictions can be applied not only to EMMO-datatypes, but any class in EMMO.
Maybe it makes sense to also include some tests about other classes e.g. materials.Atom

tests/test_restrictions_emmo.py
```diff
@@ -0,0 +1,160 @@
...
+            b_node = r._bnode
+            graph = r._graph
+
+            restriction_predicates = graph.predicates(subject=b_node)
+            quantifiers = [self.quantifiers[p] for p in restriction_predicates
+                           if p in self.quantifiers]
+            self.assertTrue(len(quantifiers) == 1, "The restriction has not "
+                                                   "exactly one quantifier.")
+            quantifier = quantifiers[0]
+            self.assertIs(quantifier, r.quantifier)
```
urbanmatthias, line 67: I very much like how these tests are very general and therefore very many restriction objects are covered. I am wondering if there should be some test that checks only one or two specific restrictions. See also the comment for the __str__ test.

urbanmatthias:
I'm aware of the CI failing. This is not your fault, but an issue of the docker image. I will try to fix it in the afternoon.

tests/test_restrictions_emmo.py
urbanmatthias: The main difficulty here is probably how to get one specific restriction objects, since the result of ontology_class.axioms is in no specific order.

tests/test_restrictions_emmo.py
kysrpex: Thank you, changed.

tests/test_restrictions_emmo.py
kysrpex: Included some more classes that have restrictions, from the materials and siunits namespace.

tests/test_restrictions_emmo.py
kysrpex: I added the class TestSpecificRestrictionsEmmo for this purpose. As you say, getting one specific restriction object is tricky. 

Therefore I opted of creating the restriction, namely a blank node and the associated triples, and then instantiating a restriction object from the blank node. The information used to create the blank node is compared against the one yielded by the restriction object when the attributes are queried.

kysrpex:
Sorry, the test is still broken. Ran the wrong test before committing.

urbanmatthias:
Is this ready for review?

urbanmatthias:
@kysrpex since this is my PR, you must review it. From my point of view it can be merged!

kysrpex:
From my point of view, it is also OK.

However, there is a more philosophical concern that I would like to discuss with you. When writing the tests, I had to use the following piece of code.

```python
def get_identifier(operand):
    if isinstance(operand, rdflib.term.Identifier) \
            or isinstance(operand, rdflib.BNode):
        operand_identifier = operand
    elif isinstance(operand, OntologyEntity):
        operand_identifier = operand.iri
    elif isinstance(operand.target, Composition):
        operand_identifier = operand._bnode
    elif isinstance(operand.target, Restriction):
        operand_identifier = operand._bnode
    else:
        raise Exception(f"Unit test is incomplete, operands of type "
                        f"{type(operand)} are not considered.")
    return operand_identifier
```

Actually, would it not make sense for restrictions and compositions to be ontology entities? This would not be immediate, as the problem is that restrictions and compositions are blank nodes, which do not have an IRI.

urbanmatthias:
Yes it could make sense that they are OntologyEntities. They're basically even OntologyClasses, just without an IRI.
For the moment I would leave it as it is, though. Mainly because we currently have that all Entities have an IRI, and I don't want to mess this up unless we have a proper (non-philosophical ;) ) reason.

kysrpex (approved):

PR #574, opened 2021-02-03
515 general rdf import

urbanmatthias:


kysrpex:
As somehow writing the tests I am also experiencing  what being a "power-user" of the code is, I would like to comment on a few things regarding the user-friendliness of the current state of the implementation.

- The user is able to load RDF files that refer to namespaces of ontologies that have have not been installed. I think we should include an exception or a warning for this case.
- A single OWL file might contain the TBox, RBox and the ABox.  When the user imports such a file, not only the ABox is imported, but also the TBox and RBox. Thus, those are imported as CUDS objects with oclass None. As a result, the test test_number_of_objects_loaded fails.
- It seems like one attribute is not being laded for one individual, it will have to be looked into (see test_attributes, which fails).
- Relationships are not being represented correctly (see test_relationships, which also fails). Indeed, when one uses the get method on a CUDS object to get the ones related to it by a certain relationship, the relationship still points to the IRI of the original individual instead of to the CUDS IRI.

As this is still a WIP, we should just have this in mind as we go on with the development.

urbanmatthias:
> * The user is able to load RDF files that refer to namespaces of ontologies that have have not been installed. I think we should include an exception or a warning for this case.

I agree

> * A single OWL file might contain the TBox, RBox and the ABox.  When the user imports such a file, not only the ABox is imported, but also the TBox and RBox. Thus, those are imported as CUDS objects with oclass None. As a result, the test test_number_of_objects_loaded fails.

I agree that this is an issue. As I said before, we assume that there is an ontology rdf file, which only contains TBox and a data rdf file, that contains ABox and RBox. But this assumption is obviously not always satisfied.

> * It seems like one attribute is not being laded for one individual, it will have to be looked into (see test_attributes, which fails).

I will take a look.

> * Relationships are not being represented correctly (see test_relationships, which also fails). Indeed, when one uses the get method on a CUDS object to get the ones related to it by a certain relationship, the relationship still points to the IRI of the original individual instead of to the CUDS IRI.

I will take a look.

> As this is still a WIP, we should just have this in mind as we go on with the development.

urbanmatthias:
Haha your tests are brilliant, there are some some cases that I completely forgot to cover 🤦🏼‍♂️

urbanmatthias:
@kysrpex can you check if the latest fixes and review?

osp/core/utils/general.py
```diff
@@ -218,31 +261,70 @@ def deserialize(json_doc, session=None, buffer_context=None,
...
+    g.parse(path, format=format)
+    test_triples = [
+        (None, rdflib.RDF.type, rdflib.OWL.Class),
+        (None, rdflib.RDF.type, rdflib.OWL.DatatypeProperty),
+        (None, rdflib.RDF.type, rdflib.OWL.ObjectProperty)
+    ]
+    if any(t in g for t in test_triples):
+        raise ValueError("Data contains class or property definitions. "
+                         "Please install ontologies using pico and use the "
+                         "rdf import only for individuals!")
```
kysrpex, line 317: Nice!

kysrpex (dismissed):

PR #576, opened 2021-02-17
552 Reduce the number of rdflib calls

kysrpex:
In Python 3.6+, a decorator @lru_cache is available in the functools module from the standard library. The idea is to apply this decorator to the OntologyEntity abstract class, after the @property decorator, for both the superclasses and the subclasses methods, so that the results of this call are cached.

A potential problem, as @urbanmatthias commented, for example, for the city ontology, each time that city.City.superclasses or from_iri is called to refer to the City ontology class, a new instance of an OntologyEntity is created, so that only for that instance the result would be cached.

Luckily, this seems not to actually affect us. I do not exactly know why, I assume it has to do with the implementation of @lru_cache. 

**Expected improvement**

_Test code_
```python
from osp.core.namespaces import city
import datetime as dt

time = dt.datetime.now()
for i in range(0, 10**4):
    city.Street.superclasses
    city.Neighborhood.superclasses
    city.City.superclasses
time = dt.datetime.now() - time
print(time)
```

Previous behavior
```
RDFLib Version: 5.0.0
0:01:31.949328
````

Current behavior, `entity_cache_size = 1024`
```
RDFLib Version: 5.0.0
0:00:11.384981
````

Current behavior, `entity_cache_size = 2`
```
RDFLib Version: 5.0.0
0:01:31.858384
````
Notice that the last one is as slow as the first one, as there is no room to cache the three results.

### CODE REVIEW

**Comment by urbanmatthias** [comment]
Wow, I'm impressed. I guess I have to spend some time learning the things provided in functools and itertools.

**Comment by pablo-de-andres** [comment]
Looks very interesting!
However, I guess this would not be persistent to multiple runs of scripts?
Wouldn't it still maybe make sense to compute them once and store them?

**Comment by urbanmatthias** [comment]
What I don't like about computing them once and storing them is:
1) The code would look uglier, since there are more files to parse
2) We would store the same info twice. Once in the rdf file and once in the precomputed file.
3) I think the performance improvement would be small: 
    parsing an extra file vs. computing subclasses only once

**Comment by pablo-de-andres** [comment]
Why an extra file?

**Comment by urbanmatthias** [comment]
How do you want to store them if not in a file?

**Comment by pablo-de-andres** [comment]
As part of the ontology rdf graph?

**Comment by urbanmatthias** [comment]
Then you loose the information about direct superclasses

**Comment by kysrpex** [comment]
If we decide to precompute this information, then we would be storing it twice, no matter whether we store it on a different file or on the RDF graph. But as @urbanmatthias said, if we put it in the RDF graph, then we loose the information about direct superclasses/subclasses. Putting the information on a new file or using a different namespace for it would solve the problem, but increase the complexity.

Also, I guess we would have only a small improvement as @urbanmatthias said if we precompute this, as we are just saving one traversal per oclass. I think the available options are either to merge this, as it improves the situation, or do a small test with very simple code loading the expected result of the query and replacing the superclasses method and subclasses method of those three subclasses directly with this expected output.

**Comment by pablo-de-andres** [comment]
Didn't consider the direct superclasses. Probably simpler then to use this (already improving) approach

**Comment by urbanmatthias** [comment]
I guess we should nevertheless also cache from_iri
```py
(osp) urba@LAPTOP-ALEQPURM:/mnt/c/Users/urba/Desktop/repos/osp-core$ /home/urba/miniconda3/envs/osp/bin/python /mnt/c/Users/urba/Desktop/repos/osp-core/test.py
0:00:06.316889
(osp) urba@LAPTOP-ALEQPURM:/mnt/c/Users/urba/Desktop/repos/osp-core$ /home/urba/miniconda3/envs/osp/bin/python /mnt/c/Users/urba/Desktop/repos/osp-core/test.py
0:00:00.498333  <-- With from_iri cached
```

urbanmatthias (approved):

PR #577, opened 2021-02-17
575 Misinterpreting default_rel: false in YAML ontology

kysrpex:
Just a little additional check was missing.


osp/core/ontology/yml/yml_parser.py
```diff
@@ -458,7 +458,8 @@ def _assert_default_relationship_occurrence(self):
         if DEFAULT_REL_KEY in self._doc:
             occurrences += 1
```
kysrpex, line 459: Also checked this part, which should be fine as it is.

urbanmatthias (approved):
Looks good

PR #578, opened 2021-02-17
Fix crash label missing

urbanmatthias:
There was a crash that was cause by an ontology not having a prefLabel for an entity. This should fix it. Tests pending.

### DIFF
--- a/osp/core/ontology/oclass.py
+++ b/osp/core/ontology/oclass.py
@@ -147,6 +147,7 @@ class OntologyClass(OntologyEntity):
                     f"commandline tool to transform entity names to CamelCase."
                 )
                 return attribute
+        raise AttributeError(name)
 
     def _get_attributes_values(self, kwargs, _force):
         """Get the cuds object's attributes from the given kwargs.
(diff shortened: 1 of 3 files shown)


osp/core/ontology/oclass.py
```diff
@@ -147,6 +147,7 @@ def get_attribute_by_argname(self, name):
                     f"commandline tool to transform entity names to CamelCase."
                 )
                 return attribute
+        raise AttributeError(name)
```
kysrpex, line 150: Makes a lot of sense.

kysrpex, line 150: Makes a lot of sense, and actually we do not have a crash for example at cuds.py:809 because we are lucky that the action performed there also raises AttributeError.

kysrpex (approved):
🎉

osp/core/ontology/oclass.py
urbanmatthias: Yes exactly. But there will be a weird Attribute error in cuds.py:809.
Something like `None does not have attribute iri`.

kysrpex (approved):

PR #581, opened 2021-02-18
580 Continuous integration fix

kysrpex:
The test B322 was designed for Python 2 and Bandit dropped support for it, therefore it no longer exists. See Bandit issue https://github.com/PyCQA/bandit/pull/662. Fixed by removing the explicit exclusion of this test when launching Bandit.


urbanmatthias:
Cool awesome!

urbanmatthias (approved):

Issue #562, opened 2021-01-14
Unable to access ontology classes whose names contain hyphen


yoavnash:
Protege allows to name an entity with a hyphen, like so: `Pre_Post-Processing_Atomistic`.
However, in the python code, this oclass will not be accessible because hyphens are not allowed according to python's syntax rules.

urbanmatthias:
I think __getitem__ should be implemented for namespaces. So maybe you can try to query it like a dictionary.

kysrpex:
Indeed, it is possible to query the namespace as a dictionary, but actually the query returns a list. We should make it return the OntologyClass directly.

```python
from osp.core.namespaces import perceptual  # From the EMMO Ontology
query_result = perceptual['1-manifold']
print(type(query_result), query_result)
```
```
<class 'list'> [<OntologyClass perceptual.1-manifold>]
```

urbanmatthias:
Ah I remember. With the square brackets, one will query by label. Since it is theoretically possible to have multiple classes with the same label, I decided to return a list.

We could assume there are no duplicate labels, but I guess there should be some fallback method for the case where there are duplicate labels.

kysrpex:
I have chosen to return just one element when using square brackets, instead of a list. Now an error will be raised when a query by label returns more than one element with the same label. The solution is almost finished, and I will commit it soon.

I like this idea, but I would like to have your opinion also on using a warning instead. I am thinking of very long simulations where all the computation time goes to waste. Maybe it would be better to show through the warning that the results may not be trustworthy and letting the user decide that, instead of forcing the process to be interrupted. Definitely I would still be against keeping the list behavior, as the results would not be as "trustworthy" as in the error on warning cases.

In the meantime, while this solution makes it to master, have you tried `getattr(namespace_object, 'Pre_Post-Processing_Atomistic')` as a workaround? This might work.

pablo-de-andres:
Would it make sense to check that there are no duplicate labels when installing the ontology?
In that case an error wouldn't be so disrupting. And we can check other constraints that we end up applying to the ontology.
Then you are sure the case won't happen

kysrpex:
> Would it make sense to check that there are no duplicate labels when installing the ontology?
> In that case an error wouldn't be so disrupting. And we can check other constraints that we end up applying to the ontology.
> Then you are sure the case won't happen

Good idea, so if two labels coincide on installation we may prevent referencing the namespace by label, and only allow suffix/IRI.

Issue #523, opened 2020-09-23
Choose label predicate in OWL ontology


urbanmatthias:
It would be nice if in e.g. osp/core/ontology/docs/emmo.yml there would be the option to specify which label predicate should be used (e.g. skos:prefLabel or rdfs:label).

Default should be rdfs:label I think.

In the same file we could also define the language of the labels.

urbanmatthias:
Hi @eich-2020 

#517 

When you want to reference an entity you either use dot notation or square brackets:

```py
<namespace>.<name>
<namespace>["<label>"]

# examples
city.Citizen
city["Citizen"]
```

Square brackets usually get the entities by a label, dot notation by URI (append <name> to URI of <namespace>).
In the case of emmo both get them by labels (reference_by_label is True).

Getting by label means the entity with the label in the square brackets is returned:

https://github.com/simphony/osp-core/blob/v3.5.0-dev/osp/core/ontology/namespace.py#L121-L123

As you can see from the link above I uses SKOS.prefLabel for that. The reason for this is that EMMO uses SKOS.prefLabel.
Some ontologies don't use SKOS.prefLabel, though.

I propose to either:
- First check if SKOS.prefLabel can be used and if not, use RDFS.label (super-property of SKOS.prefLabel, used very often).
- or let the user specify the label to use in the emmo.yml/foaf.yml file. E.g by parsing the yml and add a triple to the graph similar to https://github.com/simphony/osp-core/blob/v3.5.0-dev/osp/core/ontology/parser.py#L324-L338 .

kysrpex:
I suggest to do both actually. If both can be used then use the preference in the yml file if there is any defined, otherwise fallback to SKOS.prefLabel.

urbanmatthias:
Also an option. I updated the links in the text above.

kysrpex:
After working in the solution, I have realized that this in fact pointless to establish a preference here. What makes sense is just to check all labels regardless of whether they are RDFS or SKOS labels, and return the object. That's what I ended up implementing.

Issue #517, opened 2020-09-08
Reference by label without language tag


urbanmatthias:
When reference_by_label is True and the user executes `my_namespace.my_entity`, the system will look for the label `my_entity` in `my_namespace` with language tag `en`.  This will lead to problems if the language tags are missing or in a different language.

kysrpex:
Now the system will not care about the language tag, unless specifically instructed by using `my_namespace["my_entitity", "lang"]`. This ability to specify the language is only useful in one case: in an ontology where two different concepts have a the same name in different languages. For example "Burro" means butter in Italian but donkey in Spanish, so in that case you would refer to butter with `my_namespace["Burro", "it"]` and to donkey with `my_namespace["Burro", "es"]`.

PR #584, opened 2021-02-22
517,523,562 labels and referencing ontology entities

kysrpex:
This pull request aims to solve issues #517, #523, #562. The first commit should solve #562 and lay groundwork to solve #517 and #523.


kysrpex:
This should solve the three issues. I am writing some tests though, to check that everything works as expected.

urbanmatthias:
image

This still needs a fix, right?

kysrpex:
Yes you are right, this needs a fix.

I see that the only point of this function is to give a name to the Ontology Entity returned by the method `from_iri` in `namespace_registry.py`. 

This name seems to be more of an aesthetics matter than a functional one. In the long term, I the ideal solution may be to prefer SKOS labels to RDFS labels, and then choose a language preference according to the order system language hierarchy > english > all others sorted in alphabetical order.

However, I as they are more important issues right now, I will fix it skipping the system language hierarchy, unless that's also simple to implement.

urbanmatthias:
yes I think it is wise to skip the language hierarchy for now

urbanmatthias:
please notify me, when this is ready for review :)

kysrpex:
@urbanmatthias Check now, the changes should take care of `_get_entity_name`.

osp/core/ontology/parser.py
```diff
@@ -338,6 +342,50 @@ def _add_reference_style_triples(self, reference_styles):
...
+                            '_graph': self._graphs[identifier]})
+
+        def in_namespace(item):
+            return OntologyNamespace.__contains__(placeholder, item)
+
+        def labels_for_iri(iri):
+            return OntologyNamespace._get_labels_for_iri(placeholder, iri,
+                                                         lang=None,
+                                                         _return_literal=True)
+        # Finally check for the duplicate labels.
```
urbanmatthias, line 359:
```
labels = sorted((o, s) for s, _, o in graph.triples(None, rdflib.RDF.label, None) if s.startswith(namespace))
```
and then it should be easy to check if there are two s for the same label, isn't it?
    

```

Comment (urbanmatthias) [reply osp/core/ontology/parser.py:359]: This function could be reduced to a couple of lines, no?

File: osp/core/ontology/namespace.py [hunk 2]
```diff
@@ -85,25 +87,40 @@ def get_iri(self):
         return self._iri
```
urbanmatthias, line 88: I only had a brief look at this, but it looks nicer than before. Good work!

osp/core/ontology/namespace.py
```diff
@@ -113,86 +130,233 @@ def __getitem__(self, label):
...
+            entity_labels = self._get_labels_for_iri(iri, lang=lang)
+            if case_sensitive is False:
+                entity_labels = (label.lower() for label in entity_labels)
+                comp_label = label.lower()
+            else:
+                comp_label = label
+            if comp_label in entity_labels:
+                _name = label if self._reference_by_label else None
+                results.append(self.get_from_iri(iri, _name=_name))
+        else:
```
urbanmatthias, line 226: I'm a bit confused about these for-else stamements that you regulary use.
In this case there is no break in the loop, so couldn't you just omit the else?

osp/core/ontology/namespace.py
```diff
@@ -113,86 +130,233 @@ def __getitem__(self, label):
...
+                               f"({', '.join(element_suffixes)}) with label"
+                               f" {label} for namespace {self}."
+                               f"\n"
+                               f"Please refer to a specific element of the "
+                               f"list by calling get_from_iri(IRI) for "
+                               f"namespace {self} for one of the following "
+                               f"IRIs: " + "{iris}."
+                               .format(iris=', '.join(entity.iri for entity in
+                                                      results)))
+            else:
```
urbanmatthias, line 242: I would also remove this else. Both cases before result in an exception, so you can simply return results[0] outside of the else.

osp/core/ontology/namespace.py
kysrpex: Yes, in this case it does not make a difference. I'll omit it in a commit that I am pushing in a while.

osp/core/ontology/namespace.py
kysrpex: Also removed in future commit.

osp/core/ontology/parser.py
kysrpex: It's a bit more complicated than that because one has to consider both RDF and SKOS labels, and when two subjects have the same RDF label, but one of the items also has an SKOS label then "there is no conflict", in the sense that the preferred labeling for such item is SKOS, etc. But I am having a look at this and trying to make the whole function simpler without duplicating code from namespace.

yoavnash (commented):
I think it better to mark as a reviewer only if I should really try to review the code...

osp/core/ontology/parser.py
urbanmatthias: Ah okay, I see. Maybe you can still make it a bit simpler, then I'd be fine with merging.

kysrpex: Check the code again. I do not know if much else can be done if one wants to return all the conflicting labels in the exception, and not just the first one found. Again it feels tricky to do it without duplicating code from namespace.

urbanmatthias (approved):

osp/core/ontology/parser.py
urbanmatthias: I approved, still a small comment.

A List of tuples (like `result`) can be transformed to a tuple of lists (like `labels, iris`) using zip:

```
labels, iris = zip(*result)
```

urbanmatthias: Okay not quite, the result is a tuple of tuples:
>>> a = [(1,2), (3, 4), (5,6)]
>>> tuple(zip(*a))
((1, 3, 5), (2, 4, 6))

kysrpex: Thanks, I did not think of using it with more than two arguments. I will merge already nevertheless so that you do not have to review again.

PR #585, opened 2021-03-01
554 Parse hasValue statements

kysrpex:
This should properly handle the hasValue statements. The related [GitLab issue](https://gitlab.cc-asp.fraunhofer.de/ontology/applications/reaxpro/co2-activation/-/issues/9) seems to be fixed at least.


urbanmatthias (approved):
Looks fine to me, I'd suggest that maybe joana takes a look as well and tests it.

urbanmatthias:
@JoanaFMorgado since this is a PR that should fix one of the issues you spotted: https://gitlab.cc-asp.fraunhofer.de/ontology/applications/reaxpro/co2-activation/-/issues/9

Could you please test it if it works for you?

urbanmatthias:
It will still accept any value, but at least the default one is correct. Is this okay?
I guess the reasoner would crash if the user would decide to overwrite this default.

yoavnash (approved):

Issue #558, opened 2021-01-12
Multiple disks issue on windows and latest rdflib release.


urbanmatthias:
In rdflib, they use urllib to parse file paths. In the latest release 5.0 from April 2020 they strip off the drive letter on windows machines with multiple drives: See https://github.com/RDFLib/rdflib/pull/1172#issue-493893638
The reason for this is the use of urldecode, which confuses the drive letter with a protocol like http://

Currently the user must have the corresponding fix of rdflib installed, which is unlikely since it is not part of the latest release.

Proposed solutions:
- Catch the error tell the user to install rdflib from source. In the latest master branch the fix is available.
- Try to avoid the error even in old rdflib versions by explicitly providing the protocol file:// to rdflib's serialize. See https://github.com/RDFLib/rdflib/pull/1172#issuecomment-701703816

If possible, the second is the preferred option

pablo-de-andres:
I just ran into this issue again with @eich-2020 on his windows machine
He is using the latest OSP-core (v3.5.5) with rdflib v5.0.0. We attempted to upgrade rdflib, but OSP-core now has a constraint that it should be lower than v6, so that is not an option

kysrpex:
> I just ran into this issue again with @eich-2020 on his windows machine
> He is using the latest OSP-core (v3.5.5) with rdflib v5.0.0. We attempted to upgrade rdflib, but OSP-core now has a constraint that it should be lower than v6, so that is not an option

For reference, I paste the traceback here.

```shell
INFO 2021-08-02 12:11:47,533 [osp.core.ontology.installation]: Will install the following namespaces: ['emmo']
INFO 2021-08-02 12:11:47,539 [osp.core.ontology.namespace_registry]: Loading namespaces emmo, http://emmo.info/emmo#.
ERROR 2021-08-02 12:11:47,865 [osp.core.pico]: An Exception occurred during installation.
Traceback (most recent call last):
  File "C:\Anaconda3\lib\site-packages\osp\core\pico.py", line 65, in install_from_terminal
    ontology_installer.install(args.files)
  File "C:\Anaconda3\lib\site-packages\osp\core\ontology\installation.py", line 35, in install
    self.install(files, self.get_new_packages, False)
  File "C:\Anaconda3\lib\site-packages\osp\core\ontology\installation.py", line 151, in _install
    self.namespace_registry.load_parser(parser)
[...]

PR #588, opened 2021-03-08
558 Multiple disks issue on windows and latest rdflib release.

kysrpex:
Among the solutions proposed in issue #558, I opted for the second. Once any part of osp-core is imported, the function `rdflib.Graph.serialize` is replaced by a decorated version of itself so that it always receives `file://` at the beginning. I included a small assert statement at the beginning of each file in osp-core using the `rdflib.Graph.serialize` function and tried importing something from each of these files to check whether the patch is indeed being applied. Seems to be the case.

### DIFF
--- a/osp/__init__.py
+++ b/osp/__init__.py
@@ -3,7 +3,9 @@ __path__ = __import__('pkgutil').extend_path(__path__, __name__)
 # Patch RDFLib <= 5.0.0. See osp-core issue https://github.com/simphony/osp-core/issues/558 (the drive letter from the
 # path is stripped on Windows by the graph.Graph.serialize method of RDFLib <= 5.0.0).
 import rdflib
+import subprocess
 from urllib.parse import urlparse
+import os
 
 
 def _compare_version_leq(version, other_version):
@@ -37,9 +39,11 @@ if _compare_version_leq(rdflib.__version__, '5.0.0'):
     def graph_serialize_fix_decorator(func):
         def graph_serialize(*args, **kwargs):
             if kwargs.get('destination') is not None and not hasattr(kwargs.get('destination'), 'write'):
-                scheme, netloc, path, params, _query, fragment = urlparse('destination')
-                if scheme != 'file':
-                    kwargs['destination'] = 'file:///' + kwargs['destination']
+                scheme, netloc, path, params, _query, fragment = urlparse(kwargs['destination'])
+                if urlparse(kwargs['destination']).path == r'\\':
+                    output = subprocess.run(['mountvol', f'{scheme}:\\', '/L'], stdout=subprocess.PIPE)
+                    dos_device_path = os.path.join(output.stdout.decode().strip().replace('?', '.'), path)
+                    kwargs['destination'] = dos_device_path
             func(*args, **kwargs)
         return graph_serialize
     rdflib.Graph.serialize = graph_serialize_fix_decorator(rdflib.Graph.serialize)
--- a/osp/__init__.py
+++ b/osp/__init__.py
@@ -1,1 +1,49 @@
 __path__ = __import__('pkgutil').extend_path(__path__, __name__)
+
+# Patch RDFLib <= 5.0.0. See osp-core issue https://github.com/simphony/osp-core/issues/558 (the drive letter from the
+# path is stripped on Windows by the graph.Graph.serialize method of RDFLib <= 5.0.0).
+import rdflib
+from urllib.parse import urlparse
+
+
+def _compare_version_leq(version, other_version):
+    """Compares two software version strings.
+
+    Receives two software version strings which are just numbers separated by dots and determines whether the first one
+    is less or equal than the second one.
+
+    Args:
+        version (str): first version string (number separated by dots).
+        other_version (str) : second version string (number separated by dots).
+
+    Returns:
+        bool: whether the first version string is less or equal than the second one.
+    """
+    version = version.split('.')
+    other_version = other_version.split('.')
+    for i in range(0, min(len(version), len(other_version))):
+        if version[i] < other_version[i]:
+            return True
+        elif version[i] > other_version[i]:
+            return False
+    else:
+        if len(other_version) > len(version) and other_version[i + 1] > str(0):
+            return False
+        else:
+            return True
+
+
+if _compare_version_leq(rdflib.__version__, '5.0.0'):
+    def graph_serialize_fix_decorator(func):
+        def graph_serialize(*args, **kwargs):
+            if kwargs.get('destination') is not None and not hasattr(kwargs.get('destination'), 'write'):
+                scheme, netloc, path, params, _query, fragment = urlparse('destination')
+                if scheme != 'file':
+                    kwargs['destination'] += 'file:///'
+            func(*args, **kwargs)
+        return graph_serialize
+    rdflib.Graph.serialize = graph_serialize_fix_decorator(rdflib.Graph.serialize)
+    rdflib.Graph.serialize.patched = True
+else:
+    rdflib.Graph.serialize.patched = False
+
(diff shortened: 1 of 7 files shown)


kysrpex:
As said in the first commit message, the solution has the drawback of also reflecting the changes on RDFLib when the user imports it by him/herself. I guess this is not an issue since the only change is a bug being fixed.

I will leave the issue as a draft until tomorrow, maybe a better solution comes to my mind in the meantime.

kysrpex:
The previous commit had a mistake, and the problem turned out to be much trickier than expected. Actually I am not even sure that the fix in rdflib actually will solve the problem in newer versions, as `urllib` produces the following output.

```python
location = 'file:///C:\\Users\\user\\Downloads\\ontology.ttl'
urlparse(location)
```
```
ParseResult(scheme='file', netloc='', path='/C:\\Users\\user\\Downloads\\ontology.ttl', params='', query='', fragment='')
```

Note the leading `/` in the path. 

The approach proposed is still valid though, if standard paths, i.e. `C:\Users\user\Downloads\ontology.ttl` are converted to paths of the form the form `\\.\Volume{22e70974-0000-0000-0000-100000000000}\Users\user\Downloads\ontology.ttl`. I just committed this workaround. It is a hack, but it is what we have.

osp/__init__.py
```diff
@@ -1 +1,63 @@
...
+        def graph_serialize_fix_decorator(func):
+            def graph_serialize(*args, **kwargs):
+                if kwargs.get('destination') is not None and not hasattr(kwargs.get('destination'), 'write'):
+                    # Bug causing case.
+                    scheme, netloc, path, params, _query, fragment = urlparse(kwargs['destination'])
+                    if urlparse(kwargs['destination']).path.startswith('\\'):  # The destination is a windows path.
+                        # Call the win32 API to get the volume ID.
+                        windows_func = ctypes.windll.kernel32.GetVolumeNameForVolumeMountPointW
+                        windows_func.argtypes = wintypes.LPCWSTR, wintypes.LPWSTR, wintypes.DWORD
+                        lpszVolumeMountPoint = wintypes.LPCWSTR(f'{scheme}:\\')
```
urbanmatthias, line 52: looks a bit complicated for me. What is this? :)

urbanmatthias (commented):
It's a shame I cannot easily test this because I have only a single drive in my dev laptop...

When I wrote my solution idea I was thinking more about calling the serialize always with "file://" no matter if on windows or linux. I am also fine with decorating the serialize method like you did. Why do you need the case distiction on the platform, though? I think calling graph.serialize("file:///home/user/something") should work fine on linux as well?

urbanmatthias:
I read somewhere that __init__.py files with `__path__ = __import__('pkgutil').extend_path(__path__, __name__)` in it should contain no further statements. I cannot find the site though, and I don't remember the reasoning behind it. If we can still import wrappers and osp-core with `import osp.<...>` without anything overriding something else it should be fine IMO

osp/__init__.py
kysrpex: Yeah, that is the same thing that I though, and if you check the evolution of the code, it's what I intended to do but after trying several times I think that it is not really possible both to just wrap the serialize function and to do it this way.

The problem is the way that urllib is handling the windows paths inside the `serialize` function. The actual path needs to be on the `path` field of the ParseResult, and it does not happens with the windows path.

kysrpex:
I'll further investigate what you said about init files.

osp/__init__.py
kysrpex: About the ctypes stuff, yeah it looks very complicated also for me, but it was the only way that I found of getting windows paths by volume (which work with urlparse) and also passing the CI. At first I thought of just querying windows itself via the command line, opening a subprocess that returns a the volume ID and capturing the output, but then bandit, the security vulnerability scanner complained about it being potentially insecure (which is true). 

Thus, the solution was to query it in some other way. It turns out that python has no easy solution to call the win32 API to get this information. There are modules on pypy but I did not want to have an extra dependency just for this. ctypes on the other hand belongs to the standard library, and lets you call C functions from Python, which is precisely a way to call the win32 API to get the ID of the disk.

urbanmatthias:
Thanks for your effort. I'll test it a bit and then I'll approve!

osp/__init__.py
```diff
@@ -1 +1,60 @@
...
+        Args:
+            version (str): first version string (number separated by dots).
+            other_version (str) : second version string (number separated by dots).
+
+        Returns:
+            bool: whether the first version string is less or equal than the second one.
+        """
+        version = map(int, version.split('.'))
+        other_version = map(int, other_version.split('.'))
+        for v, o in zip(version, other_version):
```
urbanmatthias, line 29: I feel like this was too bulky and did not cover all the cases.
- You were comparing strings. Could lead to issues when comparing 2.10 with 2.2
- When everything was fine in the loop, you were only checking if the next number was larger than 0. Could lead to issues whem comparing 2.1 with 2.1.0.0.1.2.0.0 (probably not the biggest issue ;) )

urbanmatthias: I wonder if we need a unittest for this...

urbanmatthias (dismissed):
From my side it can be merged. Maybe we should add a unittest for the version comparison, but I'm not sure about that. What do you think?

osp/__init__.py
```diff
@@ -1 +1,60 @@
...
+            if v == o:
+                continue
+            return v < o
+        
+        # check remaining numbers of other_version
+        return all(o <= 0 for o in other_version)
```
urbanmatthias, line 35: since other_version is a iterator (because of `map`), here it should only iterate over the items not covered in the loop. If there are no more, `all` of an empty sequence is defined to be true.

osp/__init__.py
```diff
@@ -1 +1,60 @@
...
+        # Then patch RDFLib with the following decorator.
+        def _graph_serialize_fix_decorator(func):
+            def graph_serialize(*args, **kwargs):
+                if kwargs.get('destination') is not None and not hasattr(kwargs.get('destination'), 'write'):
+                    # Bug causing case.
+                    scheme, netloc, path, params, _query, fragment = _osp_urlparse(kwargs['destination'])
+                    if _osp_urlparse(kwargs['destination']).path.startswith('\\'):  # The destination is a windows path.
+                        # Call the win32 API to get the volume ID.
+                        windows_func = _osp_ctypes.windll.kernel32.GetVolumeNameForVolumeMountPointW
+                        windows_func.argtypes = _osp_wintypes.LPCWSTR, _osp_wintypes.LPWSTR, _osp_wintypes.DWORD
```
urbanmatthias, line 48: Is there any resource describing this. Or did you come up with this on your own?

osp/__init__.py
kysrpex: You are right, your solution is way less bulky and I did not consider the problem with the strings and the additional zeroes.

kysrpex: I'll write a test for this.

osp/__init__.py
kysrpex: The [win32 API documentation for `GetVolumeNameForVolumeMountPointW` ](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getvolumenameforvolumemountpointw) and the [ctypes library documentation](https://docs.python.org/3/library/ctypes.html). I did know about ctypes, but not about this function in the Windows API, but it was mentioned online as the way to get the disk ID.

kysrpex:
> From my side it can be merged. Maybe we should add a unittest for the version comparison, but I'm not sure about that. What do you think?

Yeah I will write a test for this, better safe than sorry

kysrpex:
@urbanmatthias Now it's ready again.

urbanmatthias (approved):

PR #589, opened 2021-03-16
Allow to change default session with with

urbanmatthias:
This will change the default session in a `with` statement:

```
with SomeWrapperSession() as session:
    # default session will be session
```

- I guess this makes the requirement that all wrapper object must be instantiated with an explicitly provided session obsolete
- This will sometimes break behavior as you can see from the tests. I hope this happens only in artificial cases like the tests

### DIFF
--- a/osp/core/session/session.py
+++ b/osp/core/session/session.py
@@ -18,19 +18,18 @@ class Session(ABC):
         self._registry = Registry()
         self.root = None
         self.graph = rdflib.Graph()
-        self._previous_session = None
 
     def __enter__(self):
         """Establish the connection to the backend."""
         from osp.core.cuds import Cuds
-        self._previous_session = Cuds._session
+        Cuds._previous_session = Cuds._session
         Cuds._session = self
         return self
 
     def __exit__(self, *args):
         """Close the connection to the backend."""
         from osp.core.cuds import Cuds
-        Cuds._session = self._previous_session
+        Cuds._session = Cuds._previous_session
         self.close()
 
     def close(self):
--- a/osp/core/session/session.py
+++ b/osp/core/session/session.py
@@ -18,13 +18,19 @@ class Session(ABC):
         self._registry = Registry()
         self.root = None
         self.graph = rdflib.Graph()
+        self._previous_session = None
 
     def __enter__(self):
         """Establish the connection to the backend."""
+        from osp.core.cuds import Cuds
+        self._previous_session = Cuds._session
+        Cuds._session = self
         return self
 
     def __exit__(self, *args):
         """Close the connection to the backend."""
+        from osp.core.cuds import Cuds
+        Cuds._session = self._previous_session
         self.close()
 
     def close(self):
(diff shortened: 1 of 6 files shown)


osp/core/session/session.py
```diff
@@ -18,13 +18,19 @@ def __init__(self):
         self._registry = Registry()
         self.root = None
         self.graph = rdflib.Graph()
+        self._previous_session = None
 
     def __enter__(self):
         """Establish the connection to the backend."""
+        from osp.core.cuds import Cuds
```
pablo-de-andres, line 25: What would happen if you had nested `with` statements?
Wouldn't this always set the previous to `CoreSession`?

urbanmatthias: You're right. The previous session should be a member of Cuds I guess

urbanmatthias: I guess unittests should be added to this

urbanmatthias:
The benefit that is already implemented:

with SomeSession() as session:
    w = ns.Wrapper(session=session)
    c = ns.City()  # this would previously be created in the core session
    w.add(c)  # this would previously copy the city
    
But the user is not interested in having objects in the core session. After the change, the City would have been created in the SomeSession

osp/core/session/session.py
urbanmatthias: okay you're not right

kysrpex:
> The benefit that is already implemented:
> 
> ```
> with SomeSession() as session:
>     w = ns.Wrapper(session=session)
>     c = ns.City()  # this would previously be created in the core session
>     w.add(c)  # this would previously copy the city
> ```
> 
> But the user is not interested in having objects in the core session. After the change, the City would have been created in the SomeSession

Ok, then it's both a performance benefit (avoiding unnecessary deepcopy) and an usability benefit (the user can edit `c` directly  and see the changes reflected). Great!

osp/core/session/session.py
kysrpex: In any case, we need an unit test for this. Maybe I can write it , that would make the code review more efective.

yoavnash:
Did you check the scenarios:
- Multiple depth-levels of 'with' statements
- Multiple 'with' statements inside one 'with' statement
?

urbanmatthias:
There is no test yet, but it should work in these scenarios

osp/core/session/session.py
kysrpex, line 25: It makes not difference in practice, but would it not make sense to just import Cuds once at the top?

kysrpex (approved):
I think it's fine and seems to work in the tests.

yoavnash (approved):

urbanmatthias:
@pablo-de-andres I merged. We should update the docs. I'll create an issue there

osp/core/session/session.py
urbanmatthias: Unfortunately, it sometimes happens that there are circular imports. I tried to put it at the top, got such an error and moved it there. Ideally, we should structure our imports s.t. no circular imports occur. I didn't check why it occured in this case.

PR #590, opened 2021-03-17
fixed unittest filetransfer

urbanmatthias:
This caused an error when building an OSP-core docker image

### DIFF
--- a/tests/test_filetransfer.py
+++ b/tests/test_filetransfer.py
@@ -375,7 +375,7 @@ class TestFiletransfer(unittest.TestCase):
                              {target[0], target[2]})
             self.assertEqual(set(os.listdir(CLIENT_DIR)),
                              {target[0], target[2]})
-            self.assertEqual(os.listdir(FILES_DIR), FILES)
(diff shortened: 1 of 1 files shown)


tests/test_filetransfer.py
```diff
@@ -375,7 +375,7 @@ def test_upload(self):
                              {target[0], target[2]})
             self.assertEqual(set(os.listdir(CLIENT_DIR)),
                              {target[0], target[2]})
-            self.assertEqual(os.listdir(FILES_DIR), FILES)
+            self.assertEqual(set(os.listdir(FILES_DIR)), set(FILES))
```
kysrpex, line 378: Is it failing because of the ordering? `self.assertItemsEqual()` might be useful in that case.

urbanmatthias: Didn't know of `self.assertItemsEqual()` :)

kysrpex:
I will approve it, since it is straightforward to see that this should not break anything.

kysrpex (approved):

PR #592, opened 2021-03-17
Added decorator to disable lazy reloads

urbanmatthias:
This might become handy for decorating e.g. a _load_from_backend. Because currently it might happen, that inside load_from_backend an expired object is accessed, which will trigger another load_from_backend. I think this recursion is confusing and should be avoided.

### DIFF
--- a/osp/core/session/wrapper_session.py
+++ b/osp/core/session/wrapper_session.py
@@ -42,11 +42,37 @@ def check_consumes_buffers(func):
         and func.does_consume_buffers
 
 
+class DisableLazyReload():
+    """Context manager that disables lazy reload of expired objects."""
+    def __init__(self):
+        """Safe the previous setting."""
+        self.prev = None
+
+    def __enter__(self):
+        """Disable the lazy reload by setting the flag to false."""
+        self.pref = WrapperSession._lazy_reload_enabled
+        WrapperSession._lazy_reload_enabled = False
+
+    def __exit__(self, type, value, traceback):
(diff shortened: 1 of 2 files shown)


urbanmatthias:
Unittests still missing...

osp/core/session/wrapper_session.py
```diff
@@ -42,12 +42,38 @@ def check_consumes_buffers(func):
         and func.does_consume_buffers
 
 
+class DisableLazyReload():
+    """Context manager that disables lazy reload of expired objects."""
+    def __init__(self):
+        """Safe the previous setting."""
+        self.prev = None
```
kysrpex, line 49: Should this be `self.pref`?

urbanmatthias: Nice spot! I think this mistake doesn't break anything and probably we could even remove the __init__ altogether.,,
But I agree, if we have it it should use the same variable name as __enter__ and __exit__....

I'll change everything to prev as an abbreviation for previous

pablo-de-andres: Also, in the docstring it should be _save_ ;)

urbanmatthias: Aaaaaaaaargh :)

urbanmatthias: Fixed as well :)

kysrpex:
We should test this with some wrapper before merging.

PR #595, opened 2021-03-17
example-rdf-import-export

urbanmatthias:


examples/example_rdf_import_export.py
```diff
@@ -0,0 +1,80 @@
...
+    with open("test.rdf") as f1:
+        with open("test2.rdf", "w") as f2:
+            for line in f1:
+                match = uuid_re.match(line)
+                if match:
+                    uid = uuid.UUID(match[2])
+                    line = line.replace(match[1], "http://city.com/"
+                                        + session._registry.get(uid).name)
+                print("\t", line, end="")
+                print(line, end="", file=f2)
```
kysrpex, line 67: I do not like this on an example, which should be as simple as possible.  I agree that usually in an RDF file you do not have UUIDs or ideally you would like to have a human-readable IRI that is self explanatory. But I would not expect users to really do this, and somehow this may mislead to the user to think that this must be done and/or that this is their task. This piece of code is not focused on showing how RDF import/export works, and actually it may be good that the user sees that OSP-core works with UUIDs.

If we go ahead in the future and allow the cuds to be identified by IRIs instead of URIs, then this would be solved. Another option could be to export CUDS as blank nodes, as this is somehow what they really are from a RDF point of view, individuals without an IRI.

examples/example_rdf_import_export.py
```diff
@@ -0,0 +1,80 @@
...
+uuid_re = re.compile(r".*(http://www\.osp-core\.com/cuds#"
+                     r"([a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}"
+                     r"-[a-z0-9]{4}-[a-z0-9]{12})).*")
+
+# Create CUDS structure
+c = branch(branch(city.City(name="Freiburg"),
+                  city.City(name="Pablo"),
+                  city.City(name="Yoav"),
+                  rel=city.hasInhabitant),
+           city.Neighborhood(name="Stühlinger"),
```
kysrpex, line 30: The umlaut is making the execution fail for me at line 75 (pretty_print). Should not happen. Something is going on with the encoding of the exported file.

kysrpex (commented):
Except for the details that I mentioned, overall I like the example, specially because you see what is happening printed, without having to put effort on going hands-on.

examples/example_rdf_import_export.py
urbanmatthias: Hmm interesting, it should be UTF-8... Maybe you're windows machine decides to use sth else, because we didn't explicitly specify the encoding?

urbanmatthias: image

examples/example_rdf_import_export.py
urbanmatthias: I agree, that this is not very nice. I thought either I create the RDF file to import by hand or I do this. I choose this, because then the user knows what to expect, because I only replace the some IRIs. If I just import any RDF file, we would have to explain that file first. 

Maybe it is enough to have a comment, that explains, that this is just for demonstration purposes? Something that the user would never have to do?
Otherwise I can also extract the RDF file after the replacement and put a comment there on how I created this file.

urbanmatthias: Interestingly, it seems like BNodes can contain UUIDs:

```
>>> x = uuid.uuid4()
>>> g = rdflib.Graph()
>>> g.add((rdflib.BNode(x), rdflib.RDF.type, rdflib.OWL.Thing))
>>> g.serialize("test.ttl", format="ttl")
>>> g.serialize("test.json", format="json-ld")
>>> g.serialize("test.xml", format="xml")
```

Unfortunately the ttl export does not contain the UUID anymore, which is a no-go IMO:

```ttl
[] a <http://www.w3.org/2002/07/owl#Thing> .
```

The other serializations work, though:

```json
[
  {
    "@id": "_:68bf8f51-a8c3-4ad5-892f-83ae8f93d9b9",
    "@type": [
      "http://www.w3.org/2002/07/owl#Thing"
    ]
  }
]
```

```xml
<?xml version="1.0" encoding="UTF-8"?>
<rdf:RDF
   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
>
  <rdf:Description rdf:nodeID="68bf8f51-a8c3-4ad5-892f-83ae8f93d9b9">
    <rdf:type rdf:resource="http://www.w3.org/2002/07/owl#Thing"/>
  </rdf:Description>
</rdf:RDF>
```

examples/example_rdf_import_export.py
```diff
@@ -0,0 +1,84 @@
...
+        print("Exported from SqliteSession")
+        for line in f:
+            print("\t", line.strip())
+
+    # Usually, RDF data does not contain UUIDs as in the current output.
+    # Replace UUIDs in the RDF file.
+    # THIS IS JUST FOR DEMONSTRATION PURPOSES.
+    # THIS ALLOWS US TO SHOW THAT OSP-CORE CAN IMPORT ANY RDF DATA,
+    # AND NOT ONLY DATA THAT WAS PREVIOUSLY EXPORTED BY OSP CORE!
+    # THIS IS NOTHING THAT YOU AS A USER WOULD EVER HAVE TO DO.
```
kysrpex, line 59: Great. This makes the point clear.

examples/example_rdf_import_export.py
kysrpex: I'll check if I can make it work on my machine.

kysrpex: The statement `open()` is using the encoding `cp1252`, while rdflib expects `utf-8`. I just set the right encoding in the `open()` statements.

kysrpex (approved):

Issue #591, opened 2021-03-17
Support IntelliSense with ontology information


pablo-de-andres:
It would be a nice feature to get autocompletion and argument information when instantiating objects from the ontology classes.
For example, the code:
```py
from osp.core.namespaces import cuba

wrapper = cuba.Wrap
```
Should suggest the autocompletion to `Wrapper`, and inform that it requires a `session` parameter.

I am not sure how easy, or even if it is possible, but I think it is definitely worth looking into at some point.

pablo-de-andres:
Possibly related, detection of the existing namespaces would also be useful.
See:
![image](https://user-images.githubusercontent.com/9430234/111434022-55406c80-86ff-11eb-96f3-8844e1539127.png)

kysrpex:
I started to work on this in the linked pull request, but later I realized that it is only possible to solve this issue in a simple way when the user is working with IPython.

When using IPython (i.e. Jupyter notebooks), some times the user will be coding when the namespace has been already imported. Therefore, the `__dir__` method of the `osp.core.namespaces` package will have been defined, and autocompletion is possible. Of course, if the user has not yet imported the namespace, then the autocompletion of the ontology entities will not be possible. Nevertheless, overall implementing this would still be a positive change, so I did it.

However, when using an IDE such as VSCode or Pycharm, the story is different. The IDE does not run the code and keep a Python kernel with the result, so it cannot use the `__dir__` method for autocompletion. So in principle, unless officially supported by the IDE or as a plug-in, this would be imposibble.

There is however a hacky "solution", which I do not like too much, but maybe @yoavnash, which is a user of OSP-core and I think needs this feature can give better input on this. There exist Python type hinting stub files (`pyi` extension), which allow a developer to supply type hinting while keeping the code in the 'py' file itself clean, so that for example, it can be kept compatible with Python2. 

The trick is, that in this file you can specify types for things that do not actually exist in the `py` file itself. Look at the picture below.

![image](https://user-images.githubusercontent.com/43052541/112331491-3f134d00-8cb9-11eb-8dd1-7750ff9cd06e.png)

It is up to the IDE to support those `pyi` files. I know that at least PyCharm supports them. I do not know about the VSCode Python plugins. When the file pictured above is included, indeed PyCharm is offering me to autocomplete both the namespace name and the `Integer` entity inside it.

[...]

PR #599, opened 2021-03-23
Add autocompletion for namespaces (IPython only)

kysrpex:
This pull request aims to solve issue #591, at least for IPython users. The idea is to modify the `__dir__` method of namespaces to include their ontology entities, so that auto completion is offered. A similar thing could maybe be done for the `namespaces.py` files, I still have to do more research on this.

To Do list to complete the pull request
- [x] Add autocompletion for Ontology Entities inside a namespace.
- [x] Add autocompletion of namespaces (when writing `osp.core.namespaces.[...]`).
- [x] Write unit tests for both cases.

### DIFF
--- a/osp/core/ontology/namespace.py
+++ b/osp/core/ontology/namespace.py
@@ -4,7 +4,7 @@
 from collections.abc import Iterable
 import rdflib
 import logging
-
+import itertools
 from osp.core.ontology.entity import OntologyEntity
 from osp.core.ontology.cuba import rdflib_cuba
 from osp.core.ontology.yml.case_insensitivity import \
@@ -30,6 +30,17 @@ class OntologyNamespace():
         self._default_rel = -1
         self._reference_by_label = \
             namespace_registry._get_reference_by_label(self._iri)
+
+    def __dir__(self):
+        """Attributes available for the OntologyNamespace class.
+
+        Returns:
+            Iterable: the available attributes, which include the methods and
+                      the ontology entities in the namespace.
+        """
+        entity_autocompletion = self._iter_labels() \
+            if self._reference_by_label else self._iter_suffixes()
+        return itertools.chain(super().__dir__(), entity_autocompletion)
 
     def __str__(self):
         """Transform the namespace to a human readable string.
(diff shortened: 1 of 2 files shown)


kysrpex:
Updated the checklist (added autocompletion for namespaces).

kysrpex:
Updated the checklist (added autocompletion for namespaces). This will only work if the user first imports the `namespaces` module, i.e. `import osp.core.namespaces as namespaces`. Then autocompletion will be available for the `namespaces` object. I think it is not possible to autocomplete the statement `from osp.core.namespaces import <...>`.

tests/test_namespace.py
```diff
@@ -515,6 +515,16 @@ def test_get_entity_name(self):
...
 
+    def test_autocompletion_ipython(self):
+        """Checks that all the expected ontology entities are in __dir__.
+
+        The check is done just for the `cuba` namespace.
+        """
+        expected = {'activeRelationship', 'passiveRelationship',
+                    'relationship', 'attribute', 'path', 'Entity', 'File',
+                    'Nothing', 'Wrapper'}
+        self.assertSetEqual(set(cuba.__dir__()) & expected, expected)
```
pablo-de-andres, line 526: Here I would call `dir(cuba)`, which is what the magic function is acting for.

osp/core/ontology/namespace.py
```diff
@@ -32,6 +32,17 @@ def __init__(self, name, namespace_registry, iri):
...
+    def __dir__(self):
+        """Attributes available for the OntologyNamespace class.
+
+        Returns:
+            Iterable: the available attributes, which include the methods and
+                      the ontology entities in the namespace.
+        """
+        entity_autocompletion = self._iter_labels() \
+            if self._reference_by_label else self._iter_suffixes()
+        return itertools.chain(super().__dir__(), entity_autocompletion)
```
pablo-de-andres, line 44: I would call `dir(super())`

osp/core/ontology/namespace.py
```diff
@@ -254,20 +265,46 @@ def _get_from_label(self, label, lang=None, case_sensitive=False):
...
-        :return: An iterator over the entities.
-        :rtype: Iterator[OntologyEntity]
+        :return: An iterator over the entity IRIs.
+        :rtype: Iterator[rdflib.URIRef]
         """
         types = [rdflib.OWL.DatatypeProperty,
                  rdflib.OWL.ObjectProperty,
                  rdflib.OWL.Class]
-        return (self._namespace_registry.from_iri(s)
+        return (s
```
pablo-de-andres, line 277: I don't think this is an Iterator, is it?

osp/core/ontology/namespace.py
```diff
@@ -254,20 +265,46 @@ def _get_from_label(self, label, lang=None, case_sensitive=False):
...
                 for t in types
                 for s, _, _ in self._graph.triples((None, rdflib.RDF.type, t))
                 if s in self)
 
+    def __iter__(self):
```
pablo-de-andres, line 282: And this doesn't look like an Iterator either

tests/test_namespace.py
kysrpex: Done, thanks!

osp/core/ontology/namespace.py
kysrpex: Done.

pablo-de-andres (approved):

PR #608, opened 2021-04-13
607 _check_duplicate_labels in ontology parser raises Keyerror

kysrpex:
Fixes a bug that makes OSP-core fail before raising the actual intended exception when duplicate labels are found during the installation of an ontology (see linked issue).

### DIFF
--- a/osp/core/ontology/parser.py
+++ b/osp/core/ontology/parser.py
@@ -370,14 +370,14 @@ class Parser():
                          in labels_for_iri(iri))
         labels, iris = tuple(result[0] for result in results),\
             tuple(result[1] for result in results)
-        coincidence_search = (i
-                              for i in range(1, len(labels))
-                              if labels[i - 1] == labels[i])
+        coincidence_search = tuple(i
+                                   for i in range(1, len(labels))
+                                   if labels[i - 1] == labels[i])
         conflicting_labels = {labels[i]: set() for i in coincidence_search}
-        for i in range(1, len(conflicting_labels)):
+        for i in coincidence_search:
             conflicting_labels[labels[i]] |= {iris[i - 1], iris[i]}
         if len(conflicting_labels) > 0:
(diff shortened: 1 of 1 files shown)


osp/core/ontology/parser.py
```diff
@@ -370,14 +370,14 @@ def labels_for_iri(iri):
...
         labels, iris = tuple(result[0] for result in results),\
             tuple(result[1] for result in results)
-        coincidence_search = (i
-                              for i in range(1, len(labels))
-                              if labels[i - 1] == labels[i])
+        coincidence_search = tuple(i
+                                   for i in range(1, len(labels))
+                                   if labels[i - 1] == labels[i])
         conflicting_labels = {labels[i]: set() for i in coincidence_search}
-        for i in range(1, len(conflicting_labels)):
```
yoavnash, line 377: Can you guess the original thought behind using `range` here to skip the first element?

kysrpex: This was written there on [26a9477](https://github.com/simphony/osp-core/pull/584/commits/26a947725ee62f119c4f4df39b0141f123634b95). I remember at that time the `unitests` package received an update (or maybe it was another one related to testing) that was causing tests to fail under certain circumstances, so I could only test with the CI. That was distracting me a lot.

Then I guess my head was still thinking of somehow the index of each element of `conflicting_labels` matching the one in `labels`, which is totally wrong. In such a case, the first item would be not necessary and the code would fail when referencing `iris[0 - 1]`.

In short, a very stupid mistake that I hope I do not repeat. I guess the tests still passed because KeyError was still being raised. It was just not the wanted KeyError.

kysrpex: If you check now `coincidence_search`, there will never be a reference to iris[-1], as there is the same `range(1, len(labels))` there.

yoavnash (approved):

PR #615, opened 2021-04-23
Fix regression in `_get` method from OntologyNamespace

kysrpex:
Fix regression in `_get` method from `OntologyNamespace`. `_reference_by_label is False` was not being taken into account to choose to get the entity from its suffix.

Regression introduced in [6d1e341](https://github.com/simphony/osp-core/commit/6d1e341d68ed39ee81684013552ca49c77ba4055).

### DIFF
--- a/osp/core/ontology/namespace.py
+++ b/osp/core/ontology/namespace.py
@@ -351,7 +351,7 @@ class OntologyNamespace():
             OntologyEntity: The ontology entity
 
         """
-        if _force_by_iri is True:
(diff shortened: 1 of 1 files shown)


kysrpex:
Notice that now the condition is the logical negation of `self._reference_by_label and not _force_by_iri`. This condition was the one being used in `_do_get` before the regression was introduced [2fd46c4](https://github.com/simphony/osp-core/commit/2fd46c4328bf47809818af50ba0f942ec909cdef).

pablo-de-andres (approved):
My code works now!

PR #616, opened 2021-04-23
600 Non-installed ontologies showing as installed

kysrpex:


osp/core/ontology/parser.py
```diff
@@ -30,13 +31,22 @@
 class Parser():
     """The parser used to parse OWL ontologies in RDF format."""
 
-    def __init__(self, graph):
+    def __init__(self, graph, parser_namespace_registry=None):
```
kysrpex, line 34: The graph is now redundant.

pablo-de-andres (commented):
I agree with your comments

pablo-de-andres (approved):
Looks great!

PR #627, opened 2021-04-30
540, 594 Allow CUDS objects to have IRIs as identifiers

kysrpex:


kysrpex:
Should be finished now, work with UUID-only CUDS and even work with some wrappers for IRI-based CUDS.

@harrytodorov Can you test that the change indeed does not break UUID-only CUDS for some wrapper(s)?

osp/core/ontology/oclass.py
```diff
@@ -287,5 +300,7 @@ def __call__(self, uid=None, session=None, _force=False, **kwargs):
             attributes=self._get_attributes_values(kwargs, _force=_force),
             oclass=self,
             session=session,
+            identifier=identifier,
```
yoavnash, line 303: Isn't better to have `id` that is shorter than `identifier`?

osp/core/cuds.py
```diff
@@ -1,106 +1,146 @@
...
     @property
-    def iri(self):
+    def iri(self) -> URIRef:
         """Get the IRI of the CUDS object."""
-        return iri_from_uid(self.uid)
+        return self.identifier if type(self.identifier) is URIRef else \
+            URIRef(CUDS_NAMESPACE_IRI + str(self.identifier))
+
+    @property
+    def identifier(self) -> Union[URIRef, UUID]:
```
yoavnash, line 102: Isn't better to have `id` that is shorter than `identifier`?

kysrpex: In general, I am not a friend of very short abbreviations in code, specially having autocompletion. In this case, it is still easy to recognize the meaning of abbreviation though. If you think it is really bad to have the longer version then I can of course change it.

osp/core/ontology/datatypes.py
```diff
@@ -2,11 +2,13 @@
 
 import uuid
 import numpy as np
-import rdflib
+from rdflib import RDF, RDFS, XSD, Literal, URIRef  # , Graph, collection
```
kysrpex, line 5: Delete the comments.

osp/core/ontology/oclass.py
```diff
@@ -271,6 +278,12 @@ def __call__(self, uid=None, session=None, _force=False, **kwargs):
         Returns:
             Cuds: The created cuds object
         """
+        # Accept strings IRI identifiers.
```
kysrpex, line 281: `# Accept strings as IRI identifiers.`

osp/core/session/transport/transport_utils.py
```diff
@@ -268,8 +268,8 @@ def get_file_cuds(obj):
     """Get the file cuds out of cuds_object, or list of cuds_objects.
 
     Args:
-        obj (Union[Cuds, UUID, List[Cuds], List[UUID], None]): The object to
-            check for fie cuds..
+        obj (Union[Cuds, UUID, URIRef, List[Cuds], List[Union[UUID, URIRef],
+             None]): The object to check for fie cuds.
```
kysrpex, line 272: Fix typo.

harrytodorov (dismissed):
I have tested the changes against a larger example simulation ([here](https://gitlab.cc-asp.fraunhofer.de/simphony/wrappers/simgromacs/-/blob/master/examples/surfactant_SDS.py)) of the SimGromacs wrapper and there seem to be no breaking changes.

yoavnash (approved):

kysrpex:
Sorry @yoavnash, it was not finished yet. Check from the description of the last two commits that we are not missing any of the additional features that we agreed to include. The dcat2 ontology is still installable.

yoavnash (approved):

PR #637, opened 2021-05-19
209 Suppport for SPARQL queries

kysrpex:
Actually there is not much to do here, as the basic elements (API) to support the SPARQL queries are already there: classes representing query results and rows in such query results, abstract methods in the `TripleStoreWrapperSession` class...

I am just doing the following:
- [x] Add support for SPARQL queries also on the Core Session.
- [x] Show a warning of slowness when performing a SPARQL query on the core session.
- [x] Raise an exception when a user tries to SPARQL any session (the exception is overridden by the implementation if it exists). 
- [x] Unit tests for this.
- [x] As the core session is not accessible to the user, add some means to perform the SPARQL query (maybe a SPARQL utils function in which a session can be specified optionally, that raises a warning with the better option (`session.sparql`) for any other session).

What I dislike about the implementation is the fact that the core session is a special case. Right now, it inherits from session, but maybe at some point this should be changed. It is actually more similar to a TripleStoreWrapperSession, as the underlying backend is rdflib's graph. Maybe the concepts of Session and WrapperSession should be joined, as in the end all sessions have a backend behind, even if it is an in-memory one like the core session.

### DIFF
--- a/osp/core/session/core_session.py
+++ b/osp/core/session/core_session.py
@@ -1,11 +1,12 @@
 """The core session used as default when no backend is connected."""
 import logging
-from .sparql_session import SparqlResult, SparqlBindingSet, SPARQLSession
+from .session import Session
+from .sparql_backend import SparqlResult, SparqlBindingSet, SPARQLBackend
 
 logger = logging.getLogger(__name__)
 
 
-class CoreSession(SPARQLSession):
+class CoreSession(Session, SPARQLBackend):
     """Core default session for all objects."""
     _warned_sparql_slow = False
 
--- a/osp/core/session/core_session.py
+++ b/osp/core/session/core_session.py
@@ -37,7 +37,7 @@ class CoreSession(Session):
         return CoreSession.CoreSessionSparqlResult(result, self)
 
     class CoreSessionSparqlResult(SparqlResult):
-        """The result of a SPARQL query of an AGraph session."""
+        """The result of a SPARQL query on the core session."""
 
         def __init__(self, query_result, session):
             """Initialize the result."""
--- a/osp/core/session/core_session.py
+++ b/osp/core/session/core_session.py
@@ -1,5 +1,6 @@
 """The core session used as default when no backend is connected."""
 from .session import Session
+from .db.triplestore_wrapper_session import SparqlResult, SparqlBindingSet
 
 
 class CoreSession(Session):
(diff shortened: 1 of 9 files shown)


kysrpex:
- Add abstract queryable session, use multiple inheritance.

kysrpex:
@pablo-de-andres Changed the inheritance scheme. About the name, what about `SPARQLBackend`? Calling something `SPARQLAble` seems weird to me.

osp/core/session/sparql_backend.py
```diff
@@ -0,0 +1,89 @@
...
+    inheritance with this abstract class.
+    """
+
+    def sparql(self, query_string):
+        """Execute the given SPARQL query on the backend.
+
+        Args:
+            query_string (): The SPARQL query as a string.
+        """
+        return self._sparql(query_string=query_string.replace(
```
pablo-de-andres, line 21: Maybe add a note for future us expaining why we need to replace the root's uid?

osp/core/utils/general.py
```diff
@@ -376,3 +377,39 @@ def get_relationships_between(subj, obj):
...
+        NotImplementedError: when the session does not support SPARQL queries.
+    """
+    from osp.core.cuds import Cuds
+    from osp.core.session.core_session import CoreSession
+    session = session or Cuds._session
+    try:
+        return session.sparql(query_string)
+    except AttributeError or NotImplementedError:
+        if isinstance(session, CoreSession):
+            raise NotImplementedError("SPARQL queries on the default session "
```
pablo-de-andres, line 410: But they _are_ supported, aren't they?

osp/core/session/core_session.py
```diff
@@ -24,3 +29,53 @@ def _notify_read(self, cuds_object):
...
+
+        def __iter__(self):
+            """Iterate the result."""
+            for row in self.result:
+                yield CoreSession.CoreSessionSparqlBindingSet(row,
+                                                              self.session)
+
+        def __len__(self):
+            """Compute the number of elements in the result."""
+            # TODO: make sure it works (unit tests).
```
pablo-de-andres, line 69: Is this TODO still applicable? Should it be tested/removed?

pablo-de-andres (commented):
I made a couple of comments, please have a look.
If you think everything is alright, I can approve!

osp/core/session/sparql_backend.py
kysrpex: Done.

osp/core/utils/general.py
kysrpex: Fixed.

pablo-de-andres (approved):

PR #657, opened 2021-06-11
Decouple the parsers actions

kysrpex:
The ontology parsers currently mix the following two operations:

- Get the graph of the ontology.
- Put the ontology in the namespace registry.

It is necessary now to get just the ontology graph (e.g. send the ontologies to AllegroGraph), and thus the code needs to be decoupled.


kysrpex:
The code was a mess before and is a mess now, but at least now a YAML ontology can be read to get its graph. Previously it was directly mixed with the other ontologies in the namespace registry.

kysrpex:
This is needed to fix the AllegroGraph wrapper pipeline, see issue [#4 on GitLab](https://gitlab.cc-asp.fraunhofer.de/simphony/wrappers/allegrograph-wrapper/-/issues/4).

kysrpex:
Closing this, replaced by #658.

PR #663, opened 2021-06-30
Fetch datatypes from Sparql-Result

### LINKED ISSUE #655: SPARQL Support: Provide a way to fetch CUDS directly instead of IRIs
One use case of SPARQL support is querying to get an _ontology individual_ (CUDS).  A SPARQL query returns just the IRI of such individual, which is not as useful as getting the CUDS object itself, where one can directly check the attributes and access related objects. 

Currently, it is possible to get the CUDS from the IRI, but it is undocumented and hard. Let's see this in more detail. With an example.

-------

Look for _person_ ontology individuals using SPARQL and get CUDS from the fetched IRIs.

```python
    persons = []
    with session.sparql(f"SELECT ?person WHERE {{"
                        f"<{c.iri}> <{city.hasInhabitant.iri}> ?person . "
                        f"}}") as result:
        for row in result:
            person = next(session.load_from_iri(row['person']))
            persons.append(person)
```

If one is working the core session, `load_from_iri` is not available to the user (because the core session is not available to the user). A workaround is to retrieve the `session = c.session`, where c is any CUDS that initialized outside of any session (they get assigned to the core session automatically).

-----

Solutions applied to the `row`s:
- Change the `sparql` function so that it automatically gives backs CUDS when detecting a cuds IRI.
- Implement a syntax such as `row['person', 'CUDS']`. Instead of `CUDS` python types could be used to get a specific representation.
- I think it may be even impossible but do python "Stunt" `row['person']` gives back the IRI or literal, `row['person'].cuds` or `row['person'].string` gives the corresponding representation. Can be mixed with the above.

Solutions applied to the `result` that affect ALL variables at once (caveat):
- Return _rows_ that give back CUDS using something like `for row in result.cuds`. A similar interface could be used to get python objects from the rdflib objects, i.e `for row in result.python`.
- Same idea as before but with this syntax `result['cuds', 'python']`. Allows mixing types.
- Same idea as before but take into account the order in which the types are specified. In the above example, OSP-core would try to convert the rows to cuds, if it fails, it will return python basic types for that variable. It's a good option for mitigating the caveat.
- Use a syntax like `for row in result(func)`, where `func` is passed by the user and handles the decision of the data type conversion. 

Solutions applied to the `result` that let the user choose the datatype for each variable:
- Use `for row in result(specimen='cuds', cycles=float)`, so that a specific datatype can be specified for each variable.
- Variant of the above: accept tuples for each column to try to convert to the first datatype first and if the conversion fails to the second.

Hybrid solution:
- Use `for row in result(specimen='cuds', cycles=float)`, so that a specific datatype can be specified for each variable, but then let the use something from the _solutions applied to the `row`s_ section to override the output. For example maybe there is a variable that is a string literal but that in some cases contains an IRI stored as a string, and in that case the user wants to use an if statement to do the check do something different with it. The `results` keyword arguments could also take tuples as an argument as in the second solution of the previous section.

MBueschelberger:
closes #655

### DIFF
--- a/osp/core/session/core_session.py
+++ b/osp/core/session/core_session.py
@@ -58,12 +58,11 @@ class CoreSession(Session, SPARQLBackend):
             """Close the connection."""
             pass
 
-        def __iter__(self, **kwargs):
+        def __iter__(self):
             """Iterate the result."""
             for row in self.result:
                 yield CoreSession.CoreSessionSparqlBindingSet(row,
-                                                              self.session,
-                                                              kwargs)
+                                                              self.session)
 
         def __len__(self):
             """Compute the number of elements in the result."""
@@ -72,10 +71,10 @@ class CoreSession(Session, SPARQLBackend):
     class CoreSessionSparqlBindingSet(SparqlBindingSet):
         """A row in the result. Mapping from variable to value."""
 
-        def __init__(self, row, session, datatypes=None):
+        def __init__(self, row, session):
             """Initialize the row."""
             self.binding_set = row
-            super().__init__(session, datatypes)
+            super().__init__(session)
 
         def _get(self, variable_name):
             return self.binding_set[variable_name]
--- a/osp/core/session/core_session.py
+++ b/osp/core/session/core_session.py
@@ -72,7 +72,7 @@ class CoreSession(Session, SPARQLBackend):
     class CoreSessionSparqlBindingSet(SparqlBindingSet):
         """A row in the result. Mapping from variable to value."""
 
-        def __init__(self, row, session, datatypes):
+        def __init__(self, row, session, datatypes=None):
             """Initialize the row."""
             self.binding_set = row
             super().__init__(session, datatypes)
--- a/osp/core/session/core_session.py
+++ b/osp/core/session/core_session.py
@@ -58,14 +58,12 @@ class CoreSession(Session, SPARQLBackend):
             """Close the connection."""
             pass
 
-        def __iter__(self):
+        def __iter__(self, **kwargs):
             """Iterate the result."""
             for row in self.result:
-                result = CoreSession.CoreSessionSparqlBindingSet(
-                    row, self.session, self.datatypes
-                )
-                self.datatypes = dict()
-                yield result
+                yield CoreSession.CoreSessionSparqlBindingSet(row,
+                                                              self.session,
+                                                              kwargs)
 
         def __len__(self):
             """Compute the number of elements in the result."""
--- a/osp/core/session/core_session.py
+++ b/osp/core/session/core_session.py
@@ -62,7 +62,8 @@ class CoreSession(Session, SPARQLBackend):
             """Iterate the result."""
             for row in self.result:
                 yield CoreSession.CoreSessionSparqlBindingSet(row,
-                                                              self.session)
+                                                              self.session,
+                                                              self.datatypes)
 
         def __len__(self):
             """Compute the number of elements in the result."""
@@ -71,10 +72,10 @@ class CoreSession(Session, SPARQLBackend):
     class CoreSessionSparqlBindingSet(SparqlBindingSet):
         """A row in the result. Mapping from variable to value."""
 
-        def __init__(self, row, session):
+        def __init__(self, row, session, datatypes):
             """Initialize the row."""
             self.binding_set = row
-            super().__init__(session)
+            super().__init__(session, datatypes)
 
         def _get(self, variable_name):
             return self.binding_set[variable_name]
--- a/osp/core/session/core_session.py
+++ b/osp/core/session/core_session.py
@@ -61,9 +61,11 @@ class CoreSession(Session, SPARQLBackend):
         def __iter__(self):
             """Iterate the result."""
             for row in self.result:
-                yield CoreSession.CoreSessionSparqlBindingSet(row,
-                                                              self.session,
-                                                              self.datatypes)
+                result = CoreSession.CoreSessionSparqlBindingSet(
+                    row, self.session, self.datatypes
+                )
+                self.datatypes = dict()
+                yield result
 
         def __len__(self):
             """Compute the number of elements in the result."""
--- a/osp/core/session/sparql_backend.py
+++ b/osp/core/session/sparql_backend.py
@@ -49,17 +49,16 @@ class SPARQLBackend(ABC):
 class SparqlResult(ABC):
     """A base class for wrapping SPARQL results of different triple stores."""
 
-    def __init__(self, session, datatypes=None):
+    def __init__(self, session):
         """Initialize the object."""
         self.session = session
-        self.datatypes = datatypes or dict()
 
     @abstractmethod
     def close(self):
         """Close the connection."""
 
     @abstractmethod
-    def __iter__(self):
+    def __iter__(self, **kwargs):
         """Iterate the result."""
 
     @abstractmethod
@@ -72,8 +71,7 @@ class SparqlResult(ABC):
 
     def __call__(self, **kwargs):
         """Add kwargs to datatypes when class is called."""
-        self.datatypes = kwargs
-        return self.__iter__()
+        return self.__iter__(**kwargs)
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         """Close the connection."""
--- a/osp/core/session/sparql_backend.py
+++ b/osp/core/session/sparql_backend.py
@@ -71,6 +71,7 @@ class SparqlResult(ABC):
         return self
 
     def __call__(self, **kwargs):
+        """Add kwargs to datatypes when class is called"""
         self.datatypes.update(kwargs)
         return self.__iter__()
 
@@ -104,7 +105,7 @@ class SparqlBindingSet(ABC):
         return self._check_datatype(variable_name, iri)
 
     def _check_datatype(self, variable_name, iri):
-        """Check if iri shall be converted to a certain datatype"""
+        """Check if iri shall be converted to a certain datatype."""
         try:
             variable_type = self.datatypes[variable_name]
             if variable_type == 'cuds':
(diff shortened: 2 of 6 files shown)


osp/core/session/core_session.py
```diff
@@ -71,10 +72,10 @@ def __len__(self):
     class CoreSessionSparqlBindingSet(SparqlBindingSet):
         """A row in the result. Mapping from variable to value."""
 
-        def __init__(self, row, session):
+        def __init__(self, row, session, datatypes):
```
kysrpex, line 75: Could you make `datatypes` a keyword argument with default value `None` here as you do in `SparqlBindingSet`? Here it's useless, but if any other code that uses the `CoreSessionSparqlBindingSet` is added, then it would be less verbose.

osp/core/session/sparql_backend.py
```diff
@@ -52,6 +52,7 @@ class SparqlResult(ABC):
     def __init__(self, session):
```
kysrpex, line 52: I would accept here also the datatypes as keyword argument with `None` as default value (and set this default if not `None` three lines below), otherwise `None` or an empty one. Good if those result objects need to be integrated with something else.

osp/core/session/sparql_backend.py
```diff
@@ -69,6 +70,11 @@ def __enter__(self):
         """Enter the with statement."""
         return self
 
+    def __call__(self, **kwargs):
+        """Add kwargs to datatypes when class is called."""
+        self.datatypes.update(kwargs)
```
kysrpex, line 75: Replace by `self.datatypes = kwargs` as we discussed, as otherwise every call would be remembered.

osp/core/session/sparql_backend.py
```diff
@@ -91,8 +98,36 @@ def __getitem__(self, variable_name):
...
+
+    def _check_datatype(self, variable_name, iri):
+        """Check if iri shall be converted to a certain datatype."""
+        if iri:
+            try:
+                variable_type = self.datatypes[variable_name]
+                if variable_type == 'cuds':
+                    cuds_query = self.session.load_from_iri(iri)
+                    return cuds_query.first()
+                elif variable_type in [float, str, int, bool]:
```
kysrpex, line 115: Elementary types are callables, this elif block can be removed.

osp/core/session/sparql_backend.py
```diff
@@ -91,8 +98,36 @@ def __getitem__(self, variable_name):
...
+                    return variable_type(iri._value)
+                elif callable(variable_type):
+                    return variable_type(iri)
+                else:
+                    raise TypeError("Variable type not understood")
+            except KeyError:
+                return iri
+            except Exception as excep:
+                raise ValueError(excep)
```
kysrpex, line 124: I like that you are raising it as `ValueError`, as yeah it's a "variable type" but probably inadequate anyway if the operation fails.

osp/core/session/sparql_backend.py
```diff
@@ -91,8 +98,36 @@ def __getitem__(self, variable_name):
...
+        else:
+            return iri
+
+
+class SparqlDataTypes(dict):
```
kysrpex, line 129: If no additional use comes to your mind, I would just use plain simple dictionary.

tests/test_sparql.py
```diff
@@ -50,7 +50,32 @@ def test_sparql(self):
                             for result in results))
         self.assertTrue(all(result['citizen_name'].toPython() == karl.name
                             for result in results))
-        results = sparql(query, session=None)
+
+        def test_function(iri):
+            value = str(iri._value)
```
kysrpex, line 55: `value = str(iri)`

tests/test_sparql.py
```diff
@@ -50,7 +50,32 @@ def test_sparql(self):
...
+        for row in results_core_session(citizen='cuds',
+                                        citizen_age=int,
+                                        citizen_name=str,
+                                        city_name=test_function):
+            self.assertTrue(row["citizen"].is_a(karl.oclass))
+            self.assertEqual(row["citizen_age"], karl.age)
+            self.assertEqual(row["citizen_name"], karl.name)
+            self.assertTrue(row["city_name"])
+
+        self.assertDictEqual(
```
kysrpex, line 70: If I have learnt something dealing with the tests, is just test the external interface of a function or class, rather than the internal. Otherwise changing the internal behaviour afterwards is a pain. So I would just remove this and be happy when the above works.

tests/test_sparql.py
```diff
@@ -50,7 +50,32 @@ def test_sparql(self):
                             for result in results))
         self.assertTrue(all(result['citizen_name'].toPython() == karl.name
                             for result in results))
-        results = sparql(query, session=None)
```
kysrpex, line 53: The part above becomes almost redundant with the new checks. You can remove it (but do your test also for the three cases, `results_none`, `results_core_session` and `results_core_session_method`).

kysrpex:
@kysrpex Note to self: before approving this we have to think of the use with the with statement.

osp/core/session/sparql_backend.py
kysrpex: @MBueschelberger I see that you changed this. I wanted to say that this was precisely fine as it was, sorry for the misunderstanding. I'll revert this to its original status.

kysrpex:
Thanks @MBueschelberger. I think this is almost ready except for one detail, that I noticed just after our conversation on Thursday. See the explanation below.

---------
Imagine I execute the following code:
```python
for citizen in result(age=int):
    print(type(citizen['age']) is int)
    break
```

Then the outcome is `True` as we want.

But if afterwards I execute this other code

```python
for citizen in result:
    print(type(citizen['age']) is int)
    break
```

As a user I would expect the outcome to be `False`, but I get `True`, because no dictionary update happened. Before merging we should solve this little problem.

------

If you have any idea just feel free to commit. I'll consider this to be my task too, so if I come up with something I may commit as well.

MBueschelberger:
> 
> 
> Thanks @MBueschelberger. I think this is almost ready except for one detail, that I noticed just after our conversation on Thursday. See the explanation below.
> 
> Imagine I execute the following code:
> 
> ```python
> for citizen in result(age=int):
>     print(type(citizen['age']) is int)
>     break
> ```
> 
> Then the outcome is `True` as we want.
> 
> But if afterwards I execute this other code
> 
> ```python
> for citizen in result:
>     print(type(citizen['age']) is int)
>     break
> ```
> 
> As a user I would expect the outcome to be `False`, but I get `True`, because no dictionary update happened. Before merging we should solve this little problem.
> 
> If you have any idea just feel free to commit. I'll consider this to be my task too, so if I come up with something I may commit as well.

Yes, I can do so.

osp/core/session/core_session.py
```diff
@@ -61,9 +61,11 @@ def close(self):
...
             """Iterate the result."""
             for row in self.result:
-                yield CoreSession.CoreSessionSparqlBindingSet(row,
-                                                              self.session,
-                                                              self.datatypes)
+                result = CoreSession.CoreSessionSparqlBindingSet(
+                    row, self.session, self.datatypes
+                )
+                self.datatypes = dict()
+                yield result
```
MBueschelberger, line 68: @kysrpex: For resolving this thread, I do not see another possiblity than reassigning an empty `dict` right after iterating the `SparqlResult`. For sure, this is not very beautiful in terms of implementation since it needs to regarded each time while defining the `__iter__`-method. 

However, would you have another idea?

MBueschelberger: @kysrpex, I just saw that this does not work if the `SparqlResult` returns more than one row (of course not - my fault). I will try to think of another solution.

osp/core/session/sparql_backend.py
```diff
@@ -72,8 +71,7 @@ def __enter__(self):
 
     def __call__(self, **kwargs):
         """Add kwargs to datatypes when class is called."""
-        self.datatypes = kwargs
-        return self.__iter__()
+        return self.__iter__(**kwargs)
```
MBueschelberger, line 74: @kysrpex: the simplest idea was just simply passing the `kwargs` to the `__iter__` function of the `SparqlResult`. This leads to the expected behaviour [you proposed](https://github.com/simphony/osp-core/pull/663#issuecomment-873980500).

kysrpex: This has the drawback of breaking existing implementations (e.g. the Allegrograph) and requiring more attention from the wrapper developer. I finally managed to implement things without requiring changes on the wrappers. See  [81a2005](https://github.com/simphony/osp-core/pull/663/commits/81a20056676caf3d537e4bf26b81b06e1d511b6e).

kysrpex:
It passes the tests, so it works, but just to see how things would look in a real use case, I tried this with the UrWerk data.

```python
[...]
        for row in result(series='cuds', literal=float):
            print(f"SPARQL result: {row['series'], row['literal']}.")
            break
        for row in result:
            print(f"SPARQL result: {row['series'], row['literal']}.")
            break    
```
```
SPARQL result: (<mechanical.Series: 7f0b79aa-0bac-4aaf-b90d-f2945c7acfaf,  AGraphSession: @0x7f9fdfad3940>, 161.011).
SPARQL result: (rdflib.term.URIRef('http://www.osp-core.com/cuds#4ae03ee7-b940-459e-9939-dd8db3c8e5c2'), rdflib.term.Literal('761.716', datatype=rdflib.term.URIRef('http://www.w3.org/2001/XMLSchema#double'))).
```
Looks nice. Thanks @MBueschelberger.

kysrpex (approved):

osp/core/session/sparql_backend.py
```diff
@@ -69,6 +69,15 @@ def __enter__(self):
         """Enter the with statement."""
         return self
 
+    def __call__(self, **kwargs):
+        """Add kwargs to datatypes when class is called."""
+        # Set the datatypes of each returned SparqlBindingSet while keeping
+        # __iter__ as an abstract method.
+        def add_datatypes(x):
```
MBueschelberger, line 76: Very nice solution! Did not think of that :)

PR #664, opened 2021-07-01
649 Name change of files at the client side – prepended with a hash

### LINKED ISSUE #649: Name change of files at the client side – prepended with a hash
Suppose the file `tensile_test_data_output.txt` belongs to `mt.text_report_file`  which is a sub-class of  `cuba.file`,  the line of code `wrapper.session.load_by_oclass(mt.text_report_file)` loads the cuds object and downloads the result file to the dowload directory at once. (Connected via [DataspaceSession](https://github.com/simphony/osp-core/blob/master/osp/wrappers/dataspace/dataspace_session.py))

But the file name gets changed to something like `5e6aed06692b4d308b8e523633f06bd0-tensile_test_data_output.txt`(prepended with a hash - maybe hash of file or uuid) causing difficulty in identifying the files at client-side. Just because of that, an extra piece of code has to be written at the client side just to remove those hashes.  Would be more convenient if we get file names without prepended hash in the results folder.

kysrpex:
Closes #649.


yoavnash:
Hi @aje08106, could you verify that it solves your issue?

yoavnash (approved):

Issue #660, opened 2021-06-16
Installing/importing ontologies from code


yoavnash:
[Update] In some situations, it could be useful to be able to install ontologies from the code itself, rather than installing them via pico.

yoavnash:
No, but rather being able to do something such as:
```python
from osp.core.ontology import install_ontology

install_ontology('my_ontology.owl')
```

yoavnash:
Is there a chance to support it before the next major version?

kysrpex:
> Is there a chance to support it before the next major version?

You are lucky today (see #761).

yoavnash:
Great! :)
What would be the command I should call?

kysrpex:
> Great! :) What would be the command I should use?

It's on #761's description. Just pass paths separated by commas to `install` or package names separated by commas to `uninstall`. If you look at the commit there is an `overwrite` boolean argument for `install` just like in normal command-line `pico`. `packages` and `namespaces` do not take arguments (just like in normal command-line pico).

yoavnash:
It's a bit hard to extract from the description. I would suggest adding a section about it somewhere in the docs.

Ktoti:
> > Great! :) What would be the command I should use?
> 
> It's on #761's description. Just pass paths separated by commas to `install` or package names separated by commas to `uninstall`. If you look at the commit there is an `overwrite` boolean argument for `install` just like in normal command-line `pico`. `packages` and `namespaces` do not take arguments (just like in normal command-line pico).

I have some questions: Where is the description? and why do we need more than just 1 path? is the path passed the yml file path? Thanks :)

kysrpex:
> > > Great! :) What would be the command I should use?
> > 
> > 
> > It's on #761's description. Just pass paths separated by commas to `install` or package names separated by commas to `uninstall`. If you look at the commit there is an `overwrite` boolean argument for `install` just like in normal command-line `pico`. `packages` and `namespaces` do not take arguments (just like in normal command-line pico).
> 
> I have some questions: Where is the description? and why do we need more than just 1 path? is the path passed the yml file path? Thanks :)

With description I mean [this](https://github.com/simphony/osp-core/pull/761#issue-1166347757) specific message. I will answer the other questions together with the answer to [this](https://github.com/simphony/osp-core/pull/761#issuecomment-1068948630) other message from you.

Issue #636, opened 2021-05-18
Redefining containers


yoavnash:
The usefulness of the container concept in CUDS, as it's used today, is arguable. On the one hand, it allows packing CUDS objects together, which is useful, for example, to ship them over the network as one unit. On the other hand, it creates the overhead of marking (in compile time) the containment relation as an "active relationship" in every ontology that the user wishes to install. Moreover, such containers are not helpful for specifying access rights since such information inflates the size of each CUDS object.

A new redefinition of this concept could bring great value to OSP-core and its users:
- A container is not a CUDS object, but a separate kind of object.
- A container allows SimPhoNy users to cluster together CUDS objects.
- This cluster must not necessarily contain only active relationships and objects can be in the container even if they are connected at all.
- Similar to the 'repository' concept in AllegroGraph or GraphDB that separate the data, but also allow "federated queries"
- Nested containers are allowed (as opposed to AllegroGraph repositories).
- Maybe allow raw files to be associated with a container 
- A container can contain also user-defined metadata (e.g., via DCAT). For example, machine name, creation date, etc.
- A container can contain information about access rights.
- Parsers, like those mentioned in #620, can create containers in a predefined structure. As an example, this is helpful for ASCII files coming from testing machines that are very similar.

In code, this could look like this:
```python
from osp.core import Container
my_container = Container(name="myContainer") 
with my_container:
    c = ns.SomeClass() # c is added automatically to my_container 
```

In this way, containers are still useful as before for packing objects together in a recursive manner, but eliminate the need for defining active relationships, and allow to add additional information, for example, for access rights, with low overhead on the CUDS objects.

the containers themselves can be stored separately from the CUDS objects, for example, in an SQL database while CUDS objects are stored in a triple store, for the sake of modularity that separates such operational information from the actual data.

yoavnash:
Notes:
- Session has a default container that could be invisible
- Unify wrapper, session, container

yoavnash:
A sustainable fix to #626

kysrpex:
This is the progress in the design process so far (I would say almost ready for implementation if not ready): https://owncloud.fraunhofer.de/index.php/f/358366720.

yoavnash:
This issue is being addressed separately as part of the discussions on the new cuds concept.

PR #700, opened 2021-08-27
Change the session and ontology architecture to base it completely on graphs, use RDFLib stores as interfaces for external software

kysrpex:
Based on PR #682's branch.

Related: #636, #660.


pablo-de-andres:
I would suggest splitting `cuds.py` into smaller files, maybe moving the helper classes outside.
I think it would help maintenance and reviewing. A file with 1700+ lines tends to be imposing.

PR #707, opened 2021-09-27
Container implementation

### LINKED ISSUE #636: Redefining containers

kysrpex:
Based on PR #700's branch. Closes #636.

- [x] Container implementation.
- [x] Change the session's method `from_identifier` so that it yields container objects when adequate.
- [x] Change the `oclass`'s method `__call__` so that it creates container objects instead of ontology individuals when adequate.
- [x] Change the wrapper so that it is both a session and a container.
- [x] Unit tests from a user's perspective.
- [x] Wrapper unit tests from a user's perspective.

In the future there should be a way to register interactive classes with type IRIs, so that the spawning of such classes is dynamic.


kysrpex:
@yoavnash Even though still a draft, I think it makes sense to have a review from the point of view of the user experience.

Please have a look specifically at `tests/test_wrapper.py`, `tests/test_container.py` since you'll see new behavior for wrappers there and the container behavior.

PR #724, opened 2021-11-12
Fix Dockerfile (`master`)

kysrpex:
* Docker image now builds (failing previously).

* Changed deprecated `MAINTAINER` to `LABEL org.opencontainers.image.authors`.

* Do not run tox when building the docker image, as GitHub's CI is already testing every commit.

----

Original PR is #723 (merge of hotfix to `dev`). This PR is the merge of the same hotfix branch to `master`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,12 @@
 FROM ubuntu:18.04
-MAINTAINER pablo.de.andres@fraunhofer.iwm.de
+LABEL org.opencontainers.image.authors="pablo.de.andres@fraunhofer.iwm.de, jose.manuel.dominguez@iwm.fraunhofer.de, yoav.nahshon@iwm.fraunhofer.de"
 
 RUN apt-get update && \
     apt-get install -y python3.7 python3-pip
-RUN python3.7 -m pip install --upgrade pip
-
-RUN ln -s /usr/bin/python3.7 /usr/bin/python & \
-    ln -s /usr/bin/pip3 /usr/bin/pip
+RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.6 1
+RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.7 2
+RUN update-alternatives --install /usr/bin/pip pip /usr/bin/pip3 2
+RUN python -m pip install --upgrade pip
 
 ADD . /simphony/osp-core
-WORKDIR /simphony/osp-core
-
-RUN pip install tox
-RUN tox -e py37
-RUN python setup.py install
+RUN pip install /simphony/osp-core
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,9 +3,9 @@ LABEL org.opencontainers.image.authors="pablo.de.andres@fraunhofer.iwm.de, jose.
 
 RUN apt-get update && \
     apt-get install -y python3.7 python3-pip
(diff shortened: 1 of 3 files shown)


pablo-de-andres:
Do we want to have all these different versions of Python installed? I feel like it might end up confusing anyone using this as a base image. 
Also I think Ubuntu18 might be getting outdated. Why not use something like Python3.8/9 as the base image, and we don't even need to worry about installing Python or pip?

kysrpex:
> Do we want to have all these different versions of Python installed? I feel like it might end up confusing anyone using this as a base image.
Ubuntu 18.04 comes with python 3.6 and python 2.7. Moreover, none of them is the default python. It is better if we use Python 3.7 because it uses rdflib 6, thus OSP-core gets the performance boost. I do not like the multiplicity either.

> Also I think Ubuntu18 might be getting outdated. Why not use something like Python3.8/9 as the base image, and we don't even need to worry about installing Python or pip?

I agree with the change. However changing the base image is a breaking change -> major release? (stuff can break on images based on the SimPhoNy docker image). To decide what to do here, we should answer the following question.

Take the [Python docker image](https://hub.docker.com/_/python) as example. There are lots of versions. The question is: when they change the base image, do they change the version number?

The answers seems to be that they use the same version number as Python, and have an additional part on the label for the base image version (e.g. `buster`, `bullseye`...).

To have this kind of sophistication we need a separate repo only for docker images, similar to [their own repo](https://github.com/docker-library/python).

Since probably we do not want to do so much work, I'd rather go for keeping the Dockerfile version tied to OSP-core version. That is, if the base image is changed a new major version needs to be released.

pablo-de-andres:
Fair enough. However, the previous image did not work. Which means anyone using it as a base must have tuned it locally.
So whatever we release now will kinda be a breaking change?

So we could use this chance to release a minimal, working one

kysrpex:
> Fair enough. However, the previous image did not work. Which means anyone using it as a base must have tuned it locally. So whatever we release now will kinda be a breaking change?
> 
> So we could use this chance to release a minimal, working one

Great argument, you convinced me.

kysrpex:
> > Fair enough. However, the previous image did not work. Which means anyone using it as a base must have tuned it locally. So whatever we release now will kinda be a breaking change?
> > So we could use this chance to release a minimal, working one
> 
> Great argument, you convinced me.

Well sorry, actually not. 

> > Fair enough. However, the previous image did not work. Which means anyone using it as a base must have tuned it locally. So whatever we release now will kinda be a breaking change?
> > So we could use this chance to release a minimal, working one
> 
> Great argument, you convinced me.

Well, actually maybe you did not... For some earlier version it must have been working... What about people who upgrade from such version to the latest one?

pablo-de-andres:
I guess it depends on what version stopped working... Because the previous `Dockerfile` was based on `miniconda`, and the one before on `Python 3.8`. Though I assume the reason why it now fails are changes on other parts of the source code, so it could have been working for previous releases, just not the latest.

I definitely see your point and I think it is the right approach in general. Here I simply think releasing a patched version and having to keep in mind that in the next major we will simplify it is unnecessary overhead.

In the end, I see both points and will approve you decision. :)

kysrpex:
@pablo-de-andres Let's go with keeping the base image as it is. I created issue #727 for the next major version.

pablo-de-andres (approved):

PR #734, opened 2021-12-08
733 Publish OSP-core in PyPI (3.x branch)

### LINKED ISSUE #733: Publish OSP-core in PyPI
### Discussed in https://github.com/simphony/osp-core/discussions/686

<div type='discussions-op-text'>

<sup>Originally posted by **yoavnash** May 20, 2021</sup>
Publish OSP-core in pypi to enable easy installation of the package.

Options for the project name:
1. `pip install simphony`
2. `pip install osp-core`
3. `pip install simphony-osp-core`

Option 3 has the advantage of being able to extend for wrappers. For example: `pip install simphony-wrapper-simlammp`.

Relevent links:
- For testing: https://test.pypi.org/
- For registering: https://pypi.org/</div>

kysrpex:
Closes #733.


kysrpex:
@pablo-de-andres @yoavnash Unfortunately, I think we cannot do the change of package name without changing the major version number.

- Users will expect things not to break, but will have nice stuff like this.
```bash
[jose@PCDEV-L009 SimPhoNy]$ pip list | grep osp
osp-core                          3.5.9
simphony-osp                      3.6.0
```
- All wrappers will break because `osp-core` is declared as a dependency.

kysrpex:
@pablo-de-andres @yoavnash A further reflection: do we want to keep the import name as `osp.core`? This is feasible even when changing the package name, but I assume users are happy to import from the same name as the package. For us, this would be `from simphony_osp import ...` and `from simphony_osp.lammps import ...` . Again this breaks everything even more.

PR #744, opened 2022-01-24
Set the default ping_timeout for the websockets socket spawned by the `CommunicationEngineClient` to `None`.

### LINKED ISSUE #743: Transport session: connection drops during long data transfers
When a data transfer is performed through the transport session, it lasts for a maximum of 20 seconds. After that, the connection on the client hangs and drops. This is an issue when large volumes of data are being transferred (e.g. files).

kysrpex:
Closes #743.

### DIFF
--- a/osp/core/session/transport/communication_engine.py
+++ b/osp/core/session/transport/communication_engine.py
@@ -162,6 +162,12 @@ class CommunicationEngineClient():
         """
         self.uri = uri
         self.kwargs = kwargs
+        # The default `ping_timeout` is 20s. The pings are not sent during a
+        # transfer. Thus, if the transfer takes more than 20s, then the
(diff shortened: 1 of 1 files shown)


pablo-de-andres:
Isn't it  a bit risky to deactivate it completely? What if something fails, and it keeps waiting forever?

kysrpex:
> Isn't it a bit risky to deactivate it completely? What if something fails, and it keeps waiting forever?

Then it keeps waiting forever (on the client). If you pull the plug, the server notices and closes its socket. Then if you connect the network again, the server returns connection reset to the client that fails too. I imagine in rare cases this may not happen and the client may get stuck forever. I agree that this is not good but imo is better than setting a maximum time for connections.

pablo-de-andres (approved):

PR #643, opened 2021-05-25
Warn when committing unreachable objects to a database wrapper.

kysrpex:
- [x] Add a warning for unreachable objects on commit.
- [ ] Add a warning for uncommited objects on close.

Issue #626, opened 2021-04-29
Unable to reach CUDS object


yoavnash:
When creating a CUDS object associated with a wrapper session, I'm unable to reach it by iteration or by export. Am I missing something?

Sample code:
```python
from osp.core.namespaces import city
from osp.wrappers.sqlite_wrapper_session import SqliteWrapperSession
from osp.core.utils import serialize_rdf_graph

uuid_to_lookup = None

with SqliteWrapperSession("test.db") as session:
    wrapper = city.CityWrapper(session=session)
    c = city.City(name="Freiburg", session=session)
    uuid_to_lookup = c.uid
    print('1', c)
    print('2', uuid_to_lookup)
    session.commit()

with SqliteWrapperSession("test.db") as session:
    wrapper = city.CityWrapper(session=session)
    for c in wrapper.iter():
        print('3', c)
    # print('4', session.load(uuid_to_lookup).one())
    serialize_rdf_graph('./test.ttl', format='turtle', session=session)
```
this produces an empty file named `test.ttl` and the following output:
```sh
1 city.City: e455f0f8-6ad2-45dc-9d83-4126b71c79d1
2 e455f0f8-6ad2-45dc-9d83-4126b71c79d1
```

When uncommenting the line `print('4', session.load(uuid_to_lookup).one())`, the following output is added:
```sh
4 city.City: e455f0f8-6ad2-45dc-9d83-4126b71c79d1
```
and the content of the file is:
```
@prefix ns1: <http://www.osp-core.com/city#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

<http://www.osp-core.com/cuds#bfbe1cd9-5b6b-4094-a976-671224fc81d6> a ns1:City ;
    ns1:coordinates "[0, 0]"^^<http://www.osp-core.com/cuba#_datatypes/VECTOR-INT-2> ;
    ns1:name "Freiburg"^^xsd:string .
```

However, I would expect that the CUDS object will be exported even without loading it explicitly.

pablo-de-andres:
Could this be because `serialize_rdf_graph` only looks at the registry when it should load everything from the backend?

yoavnash:
That's what I think, but that misleads the user

kysrpex:
> Could this be because `serialize_rdf_graph` only looks at the registry when it should load everything from the backend?

The serializer is trying to find all CUDS objects from the root of the session, the wrapper. The city was never added to the wrapper, so it is not exported.

[...]

PR #748, opened 2022-02-01
Improve the warning provided when CUDS objects are unreachable from the wrapper object and add an option to disable it

kysrpex:
Generating such a warning involves traversing the whole connected component of the graph where the wrapper object is. This is extremely expensive if the dataset is large. Therefore, the user should be able to disable it. A better solution should be provided in SimPhoNy v4. Related to #643, #626.

This PR also improves the performance of the generation of such warning (by reusing the same sets in each iteration of the `get_subtree` method).

### DIFF
--- a/osp/core/session/db/db_wrapper_session.py
+++ b/osp/core/session/db/db_wrapper_session.py
@@ -1,18 +1,19 @@
 """An abstract session containing method useful for all database backends."""
 
-from abc import abstractmethod
-from typing import Union
 import itertools
 import logging
-import rdflib
 import uuid
-from osp.core.utils.general import uid_from_iri, CUDS_IRI_PREFIX
+from abc import abstractmethod
+from typing import Union
+
+import rdflib
+
+import osp.core.warnings as warning_settings
 from osp.core.ontology.namespace_registry import namespace_registry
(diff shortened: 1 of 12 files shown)


osp/core/session/db/db_wrapper_session.py
```diff
@@ -197,3 +201,23 @@ def _unreachable_warning(self, root_obj: Union[rdflib.URIRef, uuid.UUID]):
...
+        if len(reachable) + len(unreachable) > 1000:
+            # Recommend disabling the warning for large datasets.
+            warning = "You are working with a large dataset. "
+            if not len(unreachable) > 0:
+                warning += "By default, when committing changes, " \
+                           "OSP-core looks for objects that are unreachable " \
+                           "from the wrapper object to generate a warning. " \
+                           "Generating such "
+            else:
+                warning += "Generating the above "
```
pablo-de-andres, line 213: Not sure that the if provides that much difference in terms of the text of the warning. Maybe the distinction could be clearer/removed?

pablo-de-andres (commented):
Maybe add tests to check whether the exception is raised or not based on how the flag is set?

pablo-de-andres (approved):

PR #761, opened 2022-03-11
Enable use of pico as a module

### LINKED ISSUE #660: Installing/importing ontologies from code

kysrpex:
Closes #660.

The feature is available via the functions below.

```python
from osp.core.pico import install, namespaces, packages, uninstall
```


osp/core/ontology/namespace_registry.py
```diff
@@ -109,9 +109,18 @@ def _get(self, name):
...
-        """Update the namespaces of the namespace registry."""
+    def update_namespaces(self,
+                          modules=None,
+                          remove=None):
+        """Update the namespaces of the namespace registry.
+
+        Only needed for Python 3.6.
+        """
+        modules = list(modules) if modules is not None else []
+        remove = list(remove) if remove is not None else []
```
pablo-de-andres, line 120: You could maybe simplify this by setting the default values as `[]` directly?
Also FYI, I believe converting a set to a list and then iterating is less efficient than iterating the set directly:
```py
In [1]: import timeit

In [2]:  a = [x * x for x in range(1, 400)]

In [3]: s = set(a)

In [4]: timeit.timeit("for i in s: pass", "from __main__ import s")
Out[4]: 5.616416699999718

In [5]: timeit.timeit("l = list(s)\nfor i in l: pass", "from __main__ import s")
Out[5]: 6.821328500000163
```

tests/test_installation.py
```diff
@@ -183,5 +189,79 @@ def test_pico_migrate(self):
             'city.yml', 'graph.xml', 'namespaces.txt']))
 
 
+class PicoModule(unittest.TestCase):
+    """Test the use of pico as a Python module."""
+
+    def setUp(self) -> None:
+        """Change installation path and reset the namespace registry."""
+        self._previous_installation_path = OntologyInstallationManager\
+            .get_default_installation_path()
```
pablo-de-andres, line 198: Shouldn't you get the `path` here, in case there was a custom one?

tests/test_installation.py
```diff
@@ -183,5 +189,79 @@ def test_pico_migrate(self):
...
+                    exist_ok=True)
+        OntologyInstallationManager.set_default_installation_path(
+            str(self._new_installation_path))
+        namespace_registry.clear()
+        namespace_registry.load_graph_file(
+            OntologyInstallationManager.get_default_installation_path())
+
+    def tearDown(self) -> None:
+        """Revert changes done during the execution of the `setUp` method."""
+        OntologyInstallationManager.set_default_installation_path(
```
pablo-de-andres, line 211: Same here, how do you know the default was the one used before?

osp/core/ontology/installation.py
```diff
@@ -23,12 +25,40 @@ def __init__(self, namespace_registry=None, path=None):
...
         if self.namespace_registry is None:
-            import osp.core.namespaces as namespaces
-            self.namespace_registry = namespaces._namespace_registry
+            from osp.core.ontology.namespace_registry import namespace_registry
+            self.namespace_registry = namespace_registry
+
+    @property
+    def path(self) -> str:
+        """Path where ontologies are installed."""
+        return self._path or self.get_default_installation_path()
```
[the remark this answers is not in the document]
kysrpex: The problem with doing this in the constructor is that then the default installation path cannot be changed (it would be fixed at `__init__`time).

Using a property, the default installation path is changed for every `OntologyInstallationManager` that does not have a custom path.

osp/core/ontology/namespace_registry.py
kysrpex: In general, using a mutable default argument in Python is bad practice, because if the argument is modified inside the function (or passed to any other function and then modified), then the _default_ is also modified. On the next function call, the list would be no longer empty by default. It is true that whether this will cause problems or not depends on the specific code, in the function, but again, in general if you do this then you have to be very vigilant about what you do, so the best is to avoid mutable arguments.

About the conversion, true. In fact, I do not even need a list, nor a set, so I should not be converting anything. Just an iterable works. I will change this.

tests/test_installation.py
kysrpex: When people use `osp.core.pico.install` and `osp.core.pico.uninstall`, what they expect is that the command acts on the default installation path. This is therefore the functionality I want to test here. This is also the reason why I added the `get_default_installation_path` and `set_default_installation_path` in the first place, just be able to change the path for the test. Otherwise the test would overwrite the user's ontologies. 

Non-default paths are already tested with the [other test case in the same file](https://github.com/simphony/osp-core/blob/9fbfe4dee86dfbf4494593e0cdc5cfd5194dab2e/tests/test_installation.py#L34).

tests/test_installation.py
kysrpex: No custom path given [here](https://github.com/simphony/osp-core/blob/9fbfe4dee86dfbf4494593e0cdc5cfd5194dab2e/osp/core/pico.py#L18). Pico uses such `OntologyInstallationManager`.

osp/core/ontology/installation.py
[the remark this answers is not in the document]
pablo-de-andres: You can still use a property, but rather than doing that check every time, you can do it once in the constructor?

tests/test_installation.py
pablo-de-andres: > When people use `osp.core.pico.install` and `osp.core.pico.uninstall`, what they expect is that the command acts on the default installation path

Are we sure of this?

osp/core/ontology/installation.py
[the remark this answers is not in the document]
pablo-de-andres: I just meant
```py
class OntologyInstallationManager:

    def __init__(self, namespace_registry=None, path=None):
        self.namespace_registry = namespace_registry
        self.path = path or self.get_default_installation_path()
        if self.namespace_registry is None:
            from osp.core.ontology.namespace_registry import namespace_registry
            self.namespace_registry = namespace_registry

    @property
    def path(self) -> str:
        return self._path

    @path.setter
    def path(self, value: str):
        self._path = value
```

It's a minor difference, but you avoid one check if the path property is queried often. I guess we could run into issues if we change the default path at some point, but that would anyway be a breaking change.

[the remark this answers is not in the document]
kysrpex: > I guess we could run into issues if we change the default path at some point.

This is the thing I precisely want to avoid. Honestly I don't see `pico` as a critical element in terms of performance (I don't see people having to call `pico` thousands of times per second). For me, the balance between performance and consistency in this case is on the consistency side.

Besides, how am I supposed to test this functionality if I cannot change the default path? Remember that I only introduced this for testing and that the path is fixed as soon as I import `install`, `uninstall`, `packages` or `namespaces` from `osp.core.pico`. That is more related to [this other comment](https://github.com/simphony/osp-core/pull/761/files/9fbfe4dee86dfbf4494593e0cdc5cfd5194dab2e#r825761345) anyway.

Moreover, allowing to change the default path is **not** a breaking change, but a backwards compatible change. Existing code will always work, because it never changes it. I would say that it is even a _"bugfix"_. Remember that you can set the default path already using the `OSP_ONTOLOGIES` [environment variable](https://simphony.readthedocs.io/en/dev/utils.html?highlight=OSP_ONTOLOGIES#ontology-installation-folder). I think it is a plus if somebody sets `os.environ['OSP_ONTOLOGIES_DIR'] = 'new_path'`, and it works as expected (even though this PR is not enough to achieve so though, but goes in this direction).

tests/test_installation.py
kysrpex: > > When people use `osp.core.pico.install` and `osp.core.pico.uninstall`, what they expect is that the command acts on the default installation path
> 
> Are we sure of this?

@yoavnash You asked for this feature, so please comment. From my side, I am sure of this.

pablo-de-andres (approved):

Ktoti:
> Closes #660.
> 
> The feature is available via the functions below.
> 
> ```python
> from osp.core.pico import install, namespaces, packages, uninstall
> ```

I get this error while trying to import install in jupyter notebook in DSMS : ImportError: cannot import name 'install' from 'osp.core.pico' . Does the class have dependencies?

kysrpex:
> > Closes #660.
> > The feature is available via the functions below.
> > ```python
> > from osp.core.pico import install, namespaces, packages, uninstall
> > ```
> 
> I get this error while trying to import install in jupyter notebook in DSMS : ImportError: cannot import name 'install' from 'osp.core.pico' . Does the class have dependencies?

Please, double-check that you have installed a version of OSP-core that includes these changes. Remember that the target of this pull request is the branch called `dev` rather than `master`. This means that the change is not immediately published to [`master`](https://github.com/simphony/osp-core/tree/master/), neither to [PyPI](https://pypi.org/project/osp-core) or [GitHub's releases section ](https://github.com/simphony/osp-core/releases).

Having said that, I will answer your questions from [this message](https://github.com/simphony/osp-core/issues/660#issuecomment-1068935482).

> and why do we need more than just 1 path?

You do not _need_ more than just 1 path. You _can_ use either _1 path_ `install("/home/me/emmo-mechanical-testing.yml")` or _more_ `install("/home/me/emmo-mechanical-testing.yml", "/home/me/another-ontology.yml", "/home/me/third-ontology.yml")`. This is just like using `pico install` from the command line.

> is the path passed the yml file path?

Yes

Also:
- When you uninstall, you do not use a path, you use the name you gave to the ontology package inside the `yml` file. Again also multiple names can be used. This is just like in command-line pico.
- When you call `packages()`, you get the names of all the packages you have installed as strings. Those are the same strings you get when you call `pico list` from the command line, under the section _"Packages:"_.
- When you call `namespaces()`, you get `OntologyNamespace` objects for all the installed namespaces. This is similar to the behavior in command line-pico, where you get the names of the namespaces instead, under the section _"Namespaces:"_.

In addition, I am adding this to the documentation, as @yoavnash [suggested](https://github.com/simphony/osp-core/issues/660#issuecomment-1067894270). Would you like to do the review of this addition? I think it makes sense since I thought @yoavnash was the one that needed this feature but it turns out that it is you, so nobody else will know better what is adequate.

Ktoti:
> ```python
> osp.core.pico
> ```

How can I check if the version I have is the right one and which one it is?

yoavnash:
`pip freeze` - google is your friend :)

kysrpex:
> `pip freeze` - google is your friend :)

I am afraid this will not help because the `dev` version of OSP-core has the same version number as the master version, plus here we are speaking of the commit level.

kysrpex:
> > ```python
> > osp.core.pico
> > ```
> 
> How can I check if the version I have is the right one and which one it is?

Check where Python is importing the package from.

Example on my computer (on your computer the path may be different).

```python
In [1]: import osp.core; print(osp.core.__path__)
['~/.local/lib/python3.10/site-packages/osp/core']
```

Then open `~/.local/lib/python3.10/site-packages/osp/core/pico.py` and compare it with the file on this PR. If there is any difference then you did not install a version with these changes.

kysrpex:
@Ktoti Documentation for this is now available [here](https://simphony.readthedocs.io/en/dev/utils.html#using-pico-as-a-python-module). Please note that it is still not on the main branch, and be ready to press CTRL+F5 on your web-browser if you do not see the section I am linking.

PR #766, opened 2022-04-06
Add black and isort code formatting as a pre-commit

kysrpex:
* Add black and isort to CI.
* Add pre-commit configuration file.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Flake8
-        run: flake8 . --exclude=*/__init__.py,osp/core/ontology/docs/EMMO/* --ignore=E203, W503 --docstring-convention=google
+        run: flake8 . --exclude=*/__init__.py,osp/core/ontology/docs/EMMO/* --ignore=E203,W503 --docstring-convention=google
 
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -11,12 +11,18 @@ jobs:
       - name: Flake8
         run: flake8 . --exclude=*/__init__.py,osp/core/ontology/docs/EMMO/* --ignore=W503 --docstring-convention=google
 
+      - name: isort
+        run: isort --profile black --filter-files --line-length 79 --check-only .
+
+      - black:
+        run: black --check --verbose --diff --line-length 79 .
+
```
pablo-de-andres, line 19: Here I would just run `pre-commit run --all-files`.
That way there is no need to maintain the different tasks

kysrpex: Isn't the client already doing that with the hook? I just put this to ensure that a _rogue_ developer that skips the hook fails to pass the CI.

pablo-de-andres: Yes, but if you decide to add a new hook in a later stage, you would have to update this. Instead, you can make GitHub actions run based on the config file, and the pipeline will fail if something is wrong

pablo-de-andres (approved):

kysrpex:
Closing PR in favour of #767, whose commit history is more clear.

PR #778, opened 2022-05-13
Fix wrong argument name for `export_cuds` in `example_rdf_import_export.py`

adhamhashibon:


adhamhashibon:
Hi Guys, this was not supposed to be a pull request here, but apparently working with a fork (to be safe) is still linked to main, so if you like, take a look at it, or just ignore it. In short there is an issue running the example RDF as there is an argument mismatch, its a small change in the example file.

kysrpex:
> Hi Guys, this was not supposed to be a pull request here, but apparently working with a fork (to be safe) is still linked to main, so if you like, take a look at it, or just ignore it. In short there is an issue running the example RDF as there is an argument mismatch, its a small change in the example file.

Thank you for the PR and for pointing this out, and sorry for taking so long to take action. I will merge it now.

kysrpex (dismissed):

kysrpex (approved):

PR #783, opened 2022-05-25
Support "Operations" in SimPhoNy

### LINKED ISSUE #669: Mutli-class Individuals
Support the creation of individuals that are of multiple types/classes without knowing the classes in advance.
For example, as it is done here: https://owlready2.readthedocs.io/en/latest/class.html#mutli-class-individuals.

Use case: 
In the knowledge app, software types (e.g., database, atomistic simulation, etc.) are represented via classes. Since a specific software can be of multiple types, this feature is required.

kysrpex:
Adds support for registering methods as "actions" associated to ontology classes. For example, download and upload commands for files are now implemented in this way (and in fact need to in order to fix #669 cleanly). But it would also be possible to write actions that multiply EMMO vectors, retrieve arrays from a URL, convert EMMO units, download a DCAT dataset, ...

- See [`test_files.py`](https://github.com/simphony/osp-core/blob/154f0f2f2479a0edc08c44d7a3de773ebd3ca9a2/tests/test_wrapper.py#L265) for an example of how they are used with files (lines 276, 304 are the most relevant).
- See [`file.py`](https://github.com/simphony/osp-core/blob/154f0f2f2479a0edc08c44d7a3de773ebd3ca9a2/simphony_osp/ontology/actions/file.py#L19) and [`setup.py`](https://github.com/simphony/osp-core/blob/154f0f2f2479a0edc08c44d7a3de773ebd3ca9a2/setup.py#L39) for an example of how a developer implements an action.

**Note:** The PR is not finished because the `Container` class still needs to be converted to a set of actions too.


kysrpex:
@pablo-de-andres @paulzierep @yoavnash This should be useful for integrating ontology and data. There are of course other, more sophisticated approaches such as RDFLib lexicalizers and constructors, or triples fabricated on the fly by the wrappers. I believe it's good to have this also available to be future-proof. I think the fine-details of how a user is supposed to access the actions or whether they should be implemented in a different way are a good topic for a dev meeting.

yoavnash:
I like the idea. I wonder if `actions` is descriptive enough. `operations` could be an alternative.

kysrpex:
- [x] Use the underscore convention instead of the `@action` decorator.
- [x] Offer possibility to read actions from a folder in the home directory.
- [x] Rename `actions` to `operations`.

kysrpex:
- [x] Convert the Container class to a set of actions.

yoavnash (approved):
Looks good all in all.

PR #785, opened 2022-06-03
Performance improvements

kysrpex:
* Cache `from_identifier` for T-Boxes

* Cache `from_label` for T-Boxes


kysrpex:
@yzuuang Can you verify that this improves the situation? Or should more be done? Regardless of the answer, at least the tests seem to run faster (real, wall time is the second number). We would gain something already merging this.

```python
In [70]: no_optimization
Out[70]: 
{'TestSessionAPI': (0.333, 0.368),
 'TestOntologyAPICity': (1.316, 1.35),
 'TestOntologyAPIFOAF': (1.956, 7.874),
 'TestBundledOperations': (0.077, 0.1),
 'TestToolsPico': (7.064, 7.066),
 'TestToolsGeneral': (0.098, 0.122),
 'TestToolsImportExport': (0.829, 0.874)}

In [71]: optimization
Out[71]: 
{'TestSessionAPI': (0.137, 0.168),
 'TestOntologyAPICity': (0.308, 0.335),
 'TestOntologyAPIFOAF': (0.118, 5.98),
 'TestBundledOperations': (0.06, 0.079),
 'TestToolsPico': (6.849, 6.879),
 'TestToolsGeneral': (0.029, 0.054),
 'TestToolsImportExport': (0.203, 0.242)}
```

simphony_osp/session/session.py
```diff
@@ -428,27 +436,41 @@ def from_label(
...
-                lang=lang,
-                return_prop=False,
-                return_literal=False,
+
+        identifiers_and_labels = self.iter_labels(
+            lang=lang,
+            return_prop=False,
+            return_literal=False,
+            return_identifier=True,
+        )
```
pablo-de-andres, line 445: Would adding an extra parameter to this function to directly filter by label be more efficient, or would the complexity of the function become too high?

kysrpex: This is a super-good point. I assume you are referring to the fact that I could do something similar to
```python
def from_label(self, label: str, lang: Optional[str] = None, case_sensitive: bool = False) -> FrozenSet[OntologyEntity]:
    """[...]
    
    [...]
    """

    [...]

    identifiers = self._graph.triples((None, label_property, Literal(label)))  # let's ignore the languages for simplicity
    for identifier in identifiers:
        try:
            resuts.add(self.from_identifier(identifier))
        except KeyError:
            pass

    [...]

    return frozenset(results)
```

The default, in-memory store of RDFLib is implemented as a dictionary, so the lookup time for subjects should be `O(1)` using this approach. The approach on the PR loops over all labels in the graph, so it is `O(n)`, where `n` is the number of labels in the graph. This means that **the approach that you suggest should be much faster**, at least in theory.

There is however an obvious drawback here: case-sensitivity. I do not expect the trick to work, for example, if `label == "amount"` but in the graph the triples are of the form `self._graph.triples(None, label_property, Literal("Amount"))`.  The solution to this problem is not as easy as try both "amount" and "Amount", because "amOunt" should also work. That is, if we want to use this approach, we have to compute all the combinations of possible capitalizations and do a lookup for each one. Things become harder with languages, because for each such combination also all allowed language strings have to be tried. This requires extra code and effort and if the user has installed only a few ontologies, the number of labels might be smaller than the number of capitalization and language combinations. But of course, if the number of installed ontologies grows, then at some point it starts to be worthwhile.

I think the tradeoff could be quite good for case-senstive cases, as the number of default label properties is only two (`skos:prefLabel` and `rdfs:label`). Just two checks, sounds awesome! But the language issue is still lurking in there: is it really worth also to do the check for all allowed languages (in terms of computational effort and human effort)?

The whole problem boils down to the fact that the RDFLib graph is a _"hash table"_ combining the hash of `(label, language)`, but even in the simple case of case-sensivity, what we can easily compute is just the hash of `label`.

kysrpex: To add to this: maybe the sensible thing to do is to wait and see if some use case needs a large number of ontologies, and then invest the time in developing such an approach?

EDIT: Ok, if you fix the language then it is worth it. But I guess such a  request is not very common anyways.

kysrpex: Another thing: if we had infinite time it would be interesting to contribute something to RDFLib so that it is possible to do this kind of lookup by altering the hash table structure that is employed.

pablo-de-andres: To be honest, I did not really mean that. I think it is definitely a good point if the case was fixed, but not realistic to consider all the possible spellings otherwise. Might be interesting to see if the Graph backend allows for case-insensitive queries, but I seem to remember it was something that needed to be defined on creation.

I will try to clarify what I meant in my next comment.

simphony_osp/session/session.py
```diff
@@ -428,27 +436,41 @@ def from_label(
...
+        if case_sensitive is False:
+            comp_label = label.lower()
+            identifiers_and_labels = (
+                (label.lower(), identifier)
+                for label, identifier in identifiers_and_labels
             )
```
pablo-de-andres, line 451: This will go completely once through all the labels, and later (line 461 on) again to look for the label. I feel like it would be more efficient to compare while you iterate on the first go.

Another small note, I think `if not case_sensitive` (or even `if case_sensitive == False`) is a more pythonic way to check for a false value.

pablo-de-andres: Nevermind, it was a conceptual mistake from my side (thanks for the offline clarification).

simphony_osp/session/session.py
kysrpex: Sorry for the misunderstanding then :|

simphony_osp/session/session.py
kysrpex: Then I'll assume merging is fine (not bothering to change  or discuss anything on the small note, because of its smallness).

PR #786, opened 2022-06-10
Polish tools module

kysrpex:
* General polish and revision of tools (including semantic2dot)

* semantic2dot: allow plotting namespaces, sessions and individuals on the same graph

* semantic2dot: add IPython rendering

Note that `find` is meant to replace the CUDS selection mechanism. For example, imagine that you need to move individual `freiburg` and all the _"contained"_ individuals (inhabitants, buildings, etc) in the mereotopological sense. Then first you get the list of individuals to move using `individuals = find(freiburg, rel=city.encloses)`. After that, you add them to the new session using `session.add(individuals)`.

grafik


kysrpex:
@pablo-de-andres @yoavnash Just fixed the branch for proper readability during review.

PR #789, opened 2022-06-15
Improve performance and adapt existing bechmarks to SimPhoNy v4

kysrpex:
* Introduced `lru_cache_weak` and `lru_cache_timestamp` decorators. Introduced the `take` method to take a limited number of items from an iterator.

* Applied such decorators to methods from `OntologyEntity` and `OntologyClass` in order to achieve faster performance while keeping the cache coherent when changes are done to a `Session`'s graph.

Something similar was being done already on SimPhoNy v3, but this time the solution, even though it is quite similar, seems to be more effective.

* Adapt existing benchmarks to SimPhoNy v4.

* Fix performance on various other places so that it is on par with SimPhoNy v3.

-----

The performance was evaluated using the following code snippets.

**SimPhoNy v4**

```python
from simphony_osp.namespaces import mechanical as mt
from simphony_osp.namespaces import math, metrology
from simphony_osp.namespaces import properties as prop
from time import time
from tqdm import tqdm

n = (333636 // 3)

start = time()
for i in tqdm(range(0, n)):
    bending_stiffness_average_value = 0.0312  # some numerical value
    cableSystem = mt.CableBundle()
    bendingStiffness_average = mt.BendingStiffness()
    realNumber = math.Real(hasNumericalData=bending_stiffness_average_value)
    bendingStiffness_average.connect(realNumber, rel=metrology.hasQuantityValue)
    cableSystem.connect(bendingStiffness_average, rel=prop.hasProperty)
end = time()
print("time %.2f" % (end - start))
```

**SimPhoNy v3**

```python
from osp.core.namespaces import mechanical as mt
from osp.core.namespaces import math, metrology
from osp.core.namespaces import properties as prop
from time import time
from tqdm import tqdm
 
n = (333636 // 3)
 
start = time()
for i in tqdm(range(0, n)):
    bending_stiffness_average_value = 0.0312  # some numerical value
    cableSystem = mt.CableBundle()
    bendingStiffness_average = mt.BendingStiffness()
    realNumber = math.Real(hasNumericalData=bending_stiffness_average_value)
    bendingStiffness_average.add(realNumber, rel=metrology.hasQuantityValue)
    cableSystem.add(bendingStiffness_average, rel=prop.hasProperty)
end = time()
print("time %.2f" % (end - start))
```

Estimation given by tqdm for each case:
* _SimPhoNy v4_ **without fix**: `511/111212 [00:07<26:53, 68.62it/s]`
* _SimPhoNy v4_ **with fix**: `5308/111212 [00:07<02:25, 728.48it/s]`
* _SimPhoNy v3_: `2099/111212 [00:07<06:39, 272.89it/s]`

---

**Edit**: actually the performance increased due to other factors since the _SimphoNy v3_ test, so it is now just on the same range (see picture below from continuous benchmarking system).

grafik


simphony_osp/session/session.py
```diff
@@ -46,6 +48,10 @@
 ENTITY = TypeVar("ENTITY", bound=OntologyEntity)
 
 
+RDF_type = RDF.type
+OWL_inverseOf = OWL.inverseOf
```
pablo-de-andres, line 52: Would it make sense to define all these somewhere globally and then import them?

simphony_osp/session/session.py
```diff
@@ -347,49 +353,59 @@ def from_identifier(self, identifier: Node) -> OntologyEntity:
...
-                    compatible |= {OntologyIndividual}
+                    found |= {OntologyIndividual}
+                    break
                 except KeyError:
                     pass
-            else:
-                compatible |= found
+
+        compatible = {
+            class_ for classes in compatible.values() for class_ in classes
```
pablo-de-andres: Can't you do something like `set(compatible.values())`?

simphony_osp/session/session.py
kysrpex: Honestly what would make sense is to send a PR to RDFLib fixing this (for example with a simple `lru_cache` if the namespace does not change).

simphony_osp/session/session.py
kysrpex: Then I would be making a set of sets, which is not what I want :smile:

pablo-de-andres: so `set().union(compatible.values())`? Not sure if it is much better at this point, though

simphony_osp/utils/cache.py
```diff
@@ -0,0 +1,141 @@
...
+
+from datetime import datetime
+from functools import lru_cache, wraps
+from typing import Any, Callable, Dict, Optional, Tuple
+from weakref import ref
+
+from simphony_osp.utils.other import take
+
+
+def lru_cache_weak(
```
pablo-de-andres, line 11: Are these decorators tested?

tests/benchmark_api.py
```diff
@@ -0,0 +1,655 @@
...
+
+    def _benchmark_iterate(self, iteration: int = None):
+        self.city.get(self.identifiers[iteration])
+
+    def _benchmark_tear_down(self):
+        """Restore the previous default TBox."""
+        Session.default_ontology = self.prev_default_ontology
+
+
+def benchmark_individual_get_byidentifier(benchmark):
```
pablo-de-andres, line 290:
```suggestion
def benchmark_individual_get_by_identifier(benchmark):
```

kysrpex: Dankeschön

pablo-de-andres: Minor comment, and I see it is also in other places (also `get_byrel`, for instance). Doesn't really need to be fixed.

tests/benchmark_api.py
```diff
@@ -0,0 +1,655 @@
...
+
+    def _benchmark_iterate(self, iteration: int = None):
+        self.city.get(self.iris[iteration])
+
+    def _benchmark_tear_down(self):
+        """Restore the previous default TBox."""
+        Session.default_ontology = self.prev_default_ontology
+
+
+def benchmark_individual_get_byidentifieruriref(benchmark):
```
pablo-de-andres, line 332:
```suggestion
def benchmark_individual_get_by_identifier_uriref(benchmark):
```

tests/benchmark_api.py
```diff
@@ -0,0 +1,655 @@
...
+
+    def _benchmark_iterate(self, iteration: int = None):
+        tuple(self.city.get(rel=self.city_namespace.hasInhabitant))
+
+    def _benchmark_tear_down(self):
+        """Restore the previous default TBox."""
+        Session.default_ontology = self.prev_default_ontology
+
+
+def benchmark_individual_get_byrel(benchmark):
```
pablo-de-andres, line 378:
```suggestion
def benchmark_individual_get_by_rel(benchmark):
```

tests/benchmark_api.py
```diff
@@ -0,0 +1,655 @@
...
+
+    def _benchmark_iterate(self, iteration: int = None):
+        tuple(self.city.get(oclass=self.city_namespace.Citizen))
+
+    def _benchmark_tear_down(self):
+        """Restore the previous default TBox."""
+        Session.default_ontology = self.prev_default_ontology
+
+
+def benchmark_individual_get_byoclass(benchmark):
```
pablo-de-andres, line 426:
```suggestion
def benchmark_individual_get_by_oclass(benchmark):
```

tests/benchmark_api.py
```diff
@@ -0,0 +1,655 @@
...
+
+    def _benchmark_iterate(self, iteration: int = None):
+        next(self.iterator)
+
+    def _benchmark_tear_down(self):
+        """Restore the previous default TBox."""
+        Session.default_ontology = self.prev_default_ontology
+
+
+def benchmark_individual_iter_byidentifier(benchmark):
```
pablo-de-andres, line 467: At this point maybe you can just search for `_by` and update whatever is left 😅

simphony_osp/utils/cache.py
kysrpex: They are definitely not broken (as you see they are all around the code in this PR). Should I invest the extra time in adding tests for fine details then? Or are we happy that they work? (e.g. test that memory leaks are indeed prevented).

tests/benchmark_api.py
kysrpex: Oops true, actually the initial intention was to write them like this I believe

tests/benchmark_api.py
kysrpex: Resolving due to

> Minor comment, and I see it is also in other places (also get_byrel, for instance). Doesn't really need to be fixed.

tests/benchmark_api.py
kysrpex: Resolving due to

> Minor comment, and I see it is also in other places (also get_byrel, for instance). Doesn't really need to be fixed.

tests/benchmark_api.py
kysrpex: Resolving due to

> Minor comment, and I see it is also in other places (also get_byrel, for instance). Doesn't really need to be fixed.

tests/benchmark_api.py
kysrpex: Resolving due to

> Minor comment, and I see it is also in other places (also get_byrel, for instance). Doesn't really need to be fixed.

simphony_osp/utils/cache.py
pablo-de-andres: I guess it is not a priority now, but it should probably be done at some point

simphony_osp/session/session.py
kysrpex: It still shows the intent. But more importantly, as stated on [line 352](https://github.com/simphony/osp-core/blob/13b4b5898d4a43b762f3159bf81535bdd099588c/simphony_osp/session/session.py#L352), this method is among the most frequently called in SimPhoNy, and what you suggest is 25% faster than what is already there. So in this case it is definitely better.

kysrpex: Sent commit doing this just now

pablo-de-andres (approved):

PR #790, opened 2022-06-21
Polish the wrapper API

kysrpex:
* Simple cache for the `InterfaceDriver` to mitigate slowdowns caused by high-latency calls
grafik

* Ability to lock a session for users (it does not close when leaving the context manager).

* New SessionSet class, similar to RelationshipSet, AttributeSet and AnnotationSet.

* `get` and `iter` methods for sessions so that they function somewhat similarly to the deprecated wrapper objects.

* The `compute` method of the wrappers accepts only JSON-serializable kwargs to facilitate interoperability with the remote interface.

* The use of JSON-serializable kwargs during the initialization of an interface is encouraged. Note that this is NOT needed for interoperability with the remote interface, because remote interfaces are always initialized when the user connects.

* Removed old code that was needed for the old implementation of the SQLite wrapper, and for the YAML ontologies.

* Improved handling of assigned of values to an attribute (less likely to raise exceptions, accepts a wider variety of inputs).

* Fix bugs with the Wrapper API. Add a reference `session` to the interfaces during the `populate`, `commit` and `compute` phases. Autocommit when running `compute`.

* Remove .gitmodules file.

* Update MANIFEST.in.

*  Update year in LICENSE.md.

* Take only kwargs for the interface compute method.

* Change simphony-osp version number to 4.0.0-rc1 (prepare for pre-release).

* Run tests also for Python 3.9 and 3.10.

*  Define uid method as part of the public API of OntologyEntity, but remove the public setter.

* Update dockerfile to use python:3.10, change maintainer to SimPhoNy mailing list.


kysrpex:
The contents of this PR have been split into several PRs for readability, closing.

PR #792, opened 2022-06-30
Ability to lock a session for users

kysrpex:
Ability to lock a session for users (it does not close when leaving the context manager)


simphony_osp/session/session.py
```diff
@@ -78,7 +62,12 @@ def locked(self) -> bool:
         manager and leaving the context. Useful for setting it as the
         default environment when it is not intended to close it afterwards.
         """
-        return self._lock > 0
+        return (self._lock + bool(self._user_lock)) > 0
+
+    @locked.setter
+    def locked(self, value: bool):
+        """Lock or unlock an environment."""
```
pablo-de-andres, line 86: You have a few references to this docstring
> See the docstring of `locked` for an explanation of what locking an environment means

But I feel like this does not explain much :)

simphony_osp/session/session.py
```diff
@@ -126,6 +115,40 @@ def __bool__(self) -> bool:
...
+
+        This way of locking is not meant to be used by end users, only
+        internally within SimPhoNy code, as it allows to lock the
+        environment several times, later requiring several unlocks, which is
+        unintuitive.
+
+        See the docstring of `locked` for an explanation of what locking an
+        environment means.
+        """
+        self._lock += 1
```
pablo-de-andres, line 154: This is indeed quite unintuitive. Why do you need to (un)lock multiple times?

simphony_osp/session/session.py
kysrpex: You are reading the docstring of the _setter_ of the property. Check the [docstring of the _getter_](https://github.com/simphony/osp-core/blob/1c7692944c085b4e5ae93d256a9755804d390b58/simphony_osp/session/session.py#L58) a few lines above. I did not foresee that this misunderstanding could occur :sweat_smile:.  It may also be that even that one does not explain much.

In any case you have a better explanation below. If you think it is a good example, it can be included in the docstring itself.

**No locking**

```python
session = MyWrapper()  # -> the session is now open
with session:
    pass
# -> the session is now closed
```

**Locking feature**

```python
session = MyWrapper()  # -> the session is now open
session.locked = True
with session:
    pass
# -> the session is still open
```

Of course the use case is setting the default session temporarily to create ontology individuals inside it, without having to close it at the end (for example to do stuff in another session, without having to nest another with statement).

pablo-de-andres: My bad, I was reading the diff between the branches and did not notice it was the setter. I think it is alright then :)

simphony_osp/session/session.py
kysrpex: This is only meant to be used internally within SimPhoNy, where such a mechanism is needed. It cannot be defined as `_lock` because other classes have to call it. This is why you find some lines like this

```python
# ↓ --------------------- Public API --------------------- ↓ #
```

around the code. What is enclosed inside is what is meant to be called by an end-user. Only such functions will be documented in the [API Reference](https://simphony.readthedocs.io/en/latest/api_ref.html). I am actually frustrated with the fact that I have to include these functions not meant for the end user, because they will also show up if the user runs `help(Environment)`. If you have an idea of how to solve this without breaking Python conventions (apart from constructing a whole new set of classes only for the user, which I have already thought of) I'd be very happy.

pablo-de-andres: But why is it needed, even internally?

kysrpex: The comment [here](https://github.com/simphony/osp-core/blob/1c7692944c085b4e5ae93d256a9755804d390b58/simphony_osp/interfaces/interface.py#L266) gives some insight.

> Session must not close when exiting its context manager (just in case the user enters the context manager again inside populate).

This means that we can get away with removing the internal lock if we assume the wrapper developer will be careful. I think in `populate` and `compute` there is not much value in using a context manager because the session for the base graph is already the default. Different is the case of `commit`, where `session_old`, `session_new` and `session_base` are available. 

Only if the user does stuff like this

```python
with self.session_old:
    old_attributes = {
        row[0]
        for row in sparql(f"""[...]""")
    }

with self.session_new:
    new_attributes = {
        row[0]
        for row in sparql(f"""[...]""")
    }

with self.session_base as base:  # redundant because it is already the default
    # do some logic, for example difference in amount of attributes
    difference = len(new_attributes) - len(old_attributes)

    self._engine.increase_attributes += difference
    base.get(oclass=ontology.Simulation).one()[ontology.num_attributes] = len(new_attributes)
```

It is a highly stupid example but hopefully shows the point: if we trust the developer not to use context managers in this way inside the wrapper methods then we can remove the internal lock. Otherwise we better leave it there.

pablo-de-andres: ok, let's leave it then

pablo-de-andres (approved):

PR #793, opened 2022-06-30
Fix bugs and polish the wrapper API

kysrpex:
* The `compute` method of the wrappers accepts only JSON-serializable kwargs to facilitate interoperability with the remote interface.

* The use of JSON-serializable kwargs during the initialization of an interface is encouraged. Note that this is NOT needed for interoperability with the remote interface, because remote interfaces are always initialized when the user connects.

* Removed old code that was needed for the old implementation of the SQLite wrapper, and for the YAML ontologies.

* Improved handling of assigned of values to an attribute (less likely to raise exceptions, accepts a wider variety of inputs).

* Fix bugs with the Wrapper API. Add a reference `session` to the interfaces during the `populate`, `commit` and `compute` phases. Autocommit when running `compute`.


simphony_osp/interfaces/interface.py
```diff
@@ -496,12 +507,15 @@ def commit(self) -> None:
...
         session = self.interface.session_new
+        self.interface.session = session
         try:
             session.lock()
             with session:
                 self.interface.commit()
         finally:
             session.unlock()
+            self.interface.session_new = None
+            self.interface.session = None
```
pablo-de-andres, line 542: Isn't this code repeated a few times? Maybe worth creating a function?

simphony_osp/ontology/individual.py
```diff
@@ -2067,19 +2033,23 @@ def attributes_add(
         # TODO: prevent the end result having more than one value depending on
         #  ontology cardinality restrictions and/or functional property
         #  criteria.
-        for value in filter(lambda v: v is not None, values):
+        values = set(
+            attribute.convert_to_datatype(value)
+            for value in values
+            if value is not None
+        )
+        for value in values:
```
pablo-de-andres: Won't this iterate the values twice? (Not sure if it might be the same generator issue as last time)

kysrpex: Well spotted, thanks. Just sent a commit fixing it. If the user provides repeated attributes it is should not be a problem, as the graph is a _set_ (from set theory) of triples.

simphony_osp/interfaces/interface.py
kysrpex: The answer in this case is: I am hesitant. The thing is that the three pieces of code are very similar, but not exactly repeated.

- `session_new` is different from `session_base`

- `self.interface.compute(**kwargs)` is different from `self.interface.commit()`

The point is that we might win in lines but not in readability by making a function in this specific case. These very subtle differences add verbosity to the function, and the piece of code is very "tailored" to the particular task task. I would put in a function code that not only is repeated, but code that if I had to modify in one place, I would want automatically modified somewhere else. This is not what I feel with these pieces of code. I feel that if I have to modify any of the three, they would become substantially different, and I would end up having to break the function up. In this case I prefer the readability trade-off.

pablo-de-andres (dismissed):

PR #794, opened 2022-06-30
SessionSet, `get` and `iter` for sessions

kysrpex:
* New SessionSet class, similar to RelationshipSet, AttributeSet and AnnotationSet.

* `get` and `iter` methods for sessions so that they function somewhat similarly to the deprecated wrapper objects.


simphony_osp/ontology/entity.py
```diff
@@ -188,7 +201,7 @@ def __str__(self) -> str:
         return (
             f"{self.label}"
             if hasattr(self, "label") and self.label is not None
-            else f"{self.uid}"
+            else f"{self._uid}"
```
pablo-de-andres, line 210: Why not use the property here (and in the other cases)?

If at some point you extend the getter code (e.g. to run some checks before returning the value), they would not be applied here

simphony_osp/session/session.py
```diff
@@ -157,6 +168,254 @@ def get_default_environment(cls) -> Optional[Environment]:
...
+            #  because the query can yield repeated results, investigate the
+            #  reasons.
+            yield from {
+                row[0]
+                for row in self._session.sparql(
+                    f"""
+                    SELECT ?entity WHERE {{
+                        ?entity rdf:type/rdfs:subClassOf*
+                        <{class_.iri}> .
+                    }}
```
pablo-de-andres: Can you select distinct, and then avoid the set?

simphony_osp/session/session.py
```diff
@@ -157,6 +168,254 @@ def get_default_environment(cls) -> Optional[Environment]:
...
+        return item in self._session and (
+            item.is_a(self._class_filter) if self._class_filter else True
+        )
+
+    def update(self, other: Iterable[OntologyIndividual]) -> None:
+        """Update the set with the union of itself and others."""
+        other = set(other)
+        if self._class_filter:
+            for individual in other:
+                if not individual.is_a(self._class_filter):
```
pablo-de-andres, line 251: Here you do not check if `self._class_filter` is defined (like you do in `__contains__`. Is this correct?

pablo-de-andres (commented):
Maybe some tests for the `SessionSet` would be benefitial?

simphony_osp/session/session.py
kysrpex: Great idea!

kysrpex:
> Maybe some tests for the `SessionSet` would be benefitial?

True, I'll write them next week.

simphony_osp/ontology/entity.py
kysrpex: That's a mistake, because for a while I thought of not making the UIDS part of the public API, but then I changed my mind. I forgot to change this. Thanks.

kysrpex: Done

simphony_osp/session/session.py
kysrpex: Done

kysrpex:
> > Maybe some tests for the `SessionSet` would be benefitial?
> 
> True, I'll write them next week.

Done

PR #795, opened 2022-06-30
Ability to access file contents during commit operations and of uncommitted files

kysrpex:


tests/test_api.py
```diff
@@ -1372,12 +1372,16 @@ def setUpClass(cls):
         with tempfile.NamedTemporaryFile(
             "w", suffix=".yml", delete=False
         ) as file:
-            foaf_modified: str = """
+            foaf_url = (
```
pablo-de-andres: These changes are also part of the other PR, right?

kysrpex: They are parte of al PRs actually (so that tests can run). It can be reverted later with another PR when it comes back online.

pablo-de-andres (approved):

PR #822, opened 2022-09-26
Enh/attributes in schema validation

### LINKED ISSUE #821: Add atttributes (=data properties) to schema validation for CUDS
The schema validation under `osp.core.utils.schema_validation` does not support checking validation of data properties.

It various cases, it might be useful, if those might be validated as well.

MBueschelberger:
Simple enhancement for checking of data properties (attributes) during schema validation in SimPhoNy < v4.

Closes #821

### DIFF
--- a/osp/core/utils/schema_validation.py
+++ b/osp/core/utils/schema_validation.py
@@ -107,10 +107,9 @@ def _check_cuds_object_cardinality(origin_cuds, dest_oclass, rel, constraints):
     elif type(rel_entity) == OntologyAttribute:
         # No datatype checking since this is already done when Cuds are
         # instantiated or imported from a file
-        if rel_entity in origin_cuds.get_attributes():
-            actual_cardinality = 1
-        else:
-            actual_cardinality = 0
+        actual_cardinality = (
+            1 if rel_entity in origin_cuds.get_attributes() else 0
+        )
     else:
         raise ConsistencyError(
             f"Relation '{rel}' not supported for {origin_cuds.oclass}."
--- a/osp/core/utils/schema_validation.py
+++ b/osp/core/utils/schema_validation.py
@@ -5,6 +5,7 @@ import logging
 import yaml
 
 from osp.core.namespaces import get_entity
+from osp.core.ontology import OntologyAttribute, OntologyRelationship
 
 logger = logging.getLogger(__name__)
 
@@ -97,9 +98,24 @@ def _load_data_model_from_yaml(data_model_file):
 
 def _check_cuds_object_cardinality(origin_cuds, dest_oclass, rel, constraints):
 
-    actual_cardinality = len(
-        origin_cuds.get(rel=get_entity(rel), oclass=get_entity(dest_oclass))
-    )
+    rel_entity = get_entity(rel)
+
+    if type(rel_entity) == OntologyRelationship:
+        actual_cardinality = len(
+            origin_cuds.get(rel=rel_entity, oclass=get_entity(dest_oclass))
+        )
+    elif type(rel_entity) == OntologyAttribute:
+        # No datatype checking since this is already done when Cuds are
+        # instanciated or imported from a file
+        if rel_entity in origin_cuds.get_attributes():
+            actual_cardinality = 1
+        else:
+            actual_cardinality = 0
+        # TODO: how to apply in Simphony >= v4 for multiple data properties?
+    else:
+        raise ConsistencyError(
+            f"Relation '{rel}' not supported for {origin_cuds.oclass}."
+        )
 
     min, max = _interpret_cardinality_value_from_constraints(constraints)
     if actual_cardinality < min or actual_cardinality > max:
@@ -164,9 +180,7 @@ def _get_optional_and_mandatory_subtrees(data_model_dict):
             continue
         for relationship, neighbors in relationships.items():
             for neighbor, constraints in neighbors.items():
-                min, max = _interpret_cardinality_value_from_constraints(
-                    constraints
-                )
+                min, max = _interpret_cardinality_value_from_constraints(constraints)
                 if min == 0:
                     optional_subtrees.add(neighbor)
                 if min > 0:
--- a/osp/core/utils/schema_validation.py
+++ b/osp/core/utils/schema_validation.py
@@ -106,7 +106,7 @@ def _check_cuds_object_cardinality(origin_cuds, dest_oclass, rel, constraints):
         )
     elif type(rel_entity) == OntologyAttribute:
         # No datatype checking since this is already done when Cuds are
-        # instanciated or imported from a file
+        # instantiated or imported from a file
         if rel_entity in origin_cuds.get_attributes():
             actual_cardinality = 1
         else:
--- a/osp/core/utils/schema_validation.py
+++ b/osp/core/utils/schema_validation.py
@@ -111,7 +111,6 @@ def _check_cuds_object_cardinality(origin_cuds, dest_oclass, rel, constraints):
             actual_cardinality = 1
         else:
             actual_cardinality = 0
-        # TODO: how to apply in Simphony >= v4 for multiple data properties?
     else:
         raise ConsistencyError(
             f"Relation '{rel}' not supported for {origin_cuds.oclass}."
--- a/osp/core/utils/schema_validation.py
+++ b/osp/core/utils/schema_validation.py
@@ -180,7 +180,9 @@ def _get_optional_and_mandatory_subtrees(data_model_dict):
             continue
         for relationship, neighbors in relationships.items():
             for neighbor, constraints in neighbors.items():
-                min, max = _interpret_cardinality_value_from_constraints(constraints)
+                min, max = _interpret_cardinality_value_from_constraints(
+                    constraints
+                )
                 if min == 0:
                     optional_subtrees.add(neighbor)
                 if min > 0:
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -134,6 +134,11 @@ class TestUtils(unittest.TestCase):
             "test_validation_schema_city_with_optional_subtree.yml",
         )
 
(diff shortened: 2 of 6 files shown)


kysrpex:
Thanks for providing your own solution, I will review it within the next few days.

MBueschelberger:
@kysrpex: the CI/testing currently fails because `https://xmlns.com/foaf/spec/index.rdf` for pulling the foaf-ontology is down. We might rerun it later again.

osp/core/utils/schema_validation.py
```diff
@@ -97,9 +98,24 @@ def _load_data_model_from_yaml(data_model_file):
...
-    )
+    rel_entity = get_entity(rel)
+
+    if type(rel_entity) == OntologyRelationship:
+        actual_cardinality = len(
+            origin_cuds.get(rel=rel_entity, oclass=get_entity(dest_oclass))
+        )
+    elif type(rel_entity) == OntologyAttribute:
+        # No datatype checking since this is already done when Cuds are
+        # instanciated or imported from a file
```
kysrpex:
```suggestion
        # instantiated or imported from a file
```

Please apply this suggestion (**required**).

osp/core/utils/schema_validation.py
```diff
@@ -97,9 +98,24 @@ def _load_data_model_from_yaml(data_model_file):
...
+        if rel_entity in origin_cuds.get_attributes():
+            actual_cardinality = 1
+        else:
+            actual_cardinality = 0
+        # TODO: how to apply in Simphony >= v4 for multiple data properties?
```
kysrpex:
```suggestion
```

Please apply this suggestion (**required**).

This feature is not available in SimPhoNy v4. The idea is to integrate either [SHACL](https://www.w3.org/TR/shacl/) or [OTTR](https://ottr.xyz/) instead.

tests/test_utils.py
```diff
@@ -188,6 +193,8 @@ def test_validate_tree_against_schema(self):
             schema_file_with_missing_entity,
         )
 
+        validate_tree_against_schema(c, schema_file_with_attribute)
```
kysrpex, line 196: Thanks for also adding the tests!

osp/core/utils/schema_validation.py
```diff
@@ -97,9 +98,24 @@ def _load_data_model_from_yaml(data_model_file):
...
+        actual_cardinality = len(
+            origin_cuds.get(rel=rel_entity, oclass=get_entity(dest_oclass))
+        )
+    elif type(rel_entity) == OntologyAttribute:
+        # No datatype checking since this is already done when Cuds are
+        # instanciated or imported from a file
+        if rel_entity in origin_cuds.get_attributes():
+            actual_cardinality = 1
+        else:
+            actual_cardinality = 0
```
kysrpex: Given that SimPhoNy v3 does not support multi-valued attributes, this is indeed valid.

Below you have alternative suggestions (**not required**).

"One"-liner

```suggestion
        actual_cardinality = (
            1 if rel_entity in origin_cuds.get_attributes() else 0
        )
```

A shorter but more unreadable option: `actual_cardinality = int(rel_entity in origin_cuds.get_attributes())`.

-----

Just for future reference I also leave here what happens in SimPhoNy v3 when the underlying graph contains multi-valued attributes.

```python
In [17]: list(freiburg.session.graph.triples((freiburg.iri, None, None)))  # artificially set-up
Out[17]: 
[(rdflib.term.URIRef('http://www.osp-core.com/cuds#e78978fe-0932-49ab-9223-4ce0c1182918'),
  rdflib.term.URIRef('http://www.w3.org/1999/02/22-rdf-syntax-ns#type'),
  rdflib.term.URIRef('http://www.osp-core.com/city#City')),
 (rdflib.term.URIRef('http://www.osp-core.com/cuds#e78978fe-0932-49ab-9223-4ce0c1182918'),
  rdflib.term.URIRef('http://www.osp-core.com/city#coordinates'),
  rdflib.term.Literal('[0 0]', datatype=rdflib.term.URIRef('http://www.osp-core.com/cuba#_datatypes/VECTOR-INT-2'))),
 (rdflib.term.URIRef('http://www.osp-core.com/cuds#e78978fe-0932-49ab-9223-4ce0c1182918'),
  rdflib.term.URIRef('http://www.osp-core.com/city#name'),
  rdflib.term.Literal('Freiburg')),
 (rdflib.term.URIRef('http://www.osp-core.com/cuds#e78978fe-0932-49ab-9223-4ce0c1182918'),
  rdflib.term.URIRef('http://www.osp-core.com/city#name'),
  rdflib.term.Literal('Strasbourg'))]

In [18]: freiburg.name
Out[18]: 'Freiburg'

In [19]: freiburg.get_attributes()
Out[19]: 
{<OntologyAttribute city.coordinates>: array([0, 0]),
 <OntologyAttribute city.name>: 'Strasbourg'}
```

kysrpex (changes requested):
Thank you for submitting the PR. It is good.

I requested changes only due to two minor details: a typo and deleting a TODO (see reasons on review comments). I provided the changes as "code review suggestions" so I think you can just click a button below the suggestion to apply the changes directly on GitHub.

kysrpex:
> Thank you for submitting the PR. It is good.
> 
> I requested changes only due to two minor details: a typo and deleting a TODO (see reasons on review comments). I provided the changes as "code review suggestions" so I think you can just click a button below the suggestion to apply the changes directly on GitHub.

@MBueschelberger I will take care of the failing CI once you commit the requested changes.

kysrpex (approved):

MBueschelberger:
@kysrpex:

I was also asked for the functionality to check the actual values of the Cuds.

For my opion this makes absolutely sense if you want to write a unittest for your function which is generating Cuds with actual data.

However, my suggestion would be that this might also overwrite the `cardinality` with a `value`-keyword in the case of data properties:

*Restriction to length*
```
    city.Neighborhood:
        city.name:
            STRING:
                value: 1+
```   

Following a similar pattern as the `cardinality`, this `value` might e.g. in case of a string restrict the length (see example above) or the actual value:

*Restriction of value*
```
    city.Neighborhood:
        city.name:
            STRING:
                value: Stühlinger
```  

In the case of floats, bools and ints, there is no restriction of the length needed, but rather a statement towards the data range:

*Restriction of range*
```
    city.Citizen:
        city.age:
            INTEGER:
                value: 0+
```  
... or the actual value:

*Restriction of value*
```
    city.Citizen:
        city.age:
            INTEGER:
                value: 29
        city.name:
            STRING:
                value: Matthias
```  

In case of vectors, length, data range and value restriction might be a bit more complicated. But I think this is not a huge priority to be implemented since we are ontologizing in the EMMO-way, without any vectors as data properties.

Since you already closed this PR and deleted the branch, I can come up with a new branch with this idea.

What do you think?

kysrpex:
> @kysrpex:
> 
> I was also asked for the functionality to check the actual values of the Cuds.
> 
> For my opion this makes absolutely sense if you want to write a unittest for your function which is generating Cuds with actual data.
> 
> However, my suggestion would be that this might also overwrite the `cardinality` with a `value`-keyword in the case of data properties:
> 
> _Restriction to length_
> 
> ```
>     city.Neighborhood:
>         city.name:
>             STRING:
>                 value: 1+
> ```
> 
> Following a similar pattern as the `cardinality`, this `value` might e.g. in case of a string restrict the length (see example above) or the actual value:
> 
> _Restriction of value_
> 
> ```
>     city.Neighborhood:
>         city.name:
>             STRING:
>                 value: Stühlinger
> ```
> 
> In the case of floats, bools and ints, there is no restriction of the length needed, but rather a statement towards the data range:
> 
> _Restriction of range_
> 
> ```
>     city.Citizen:
>         city.age:
>             INTEGER:
>                 value: 0+
> ```
> 
> ... or the actual value:
> 
> _Restriction of value_
> 
> ```
>     city.Citizen:
>         city.age:
>             INTEGER:
>                 value: 29
>         city.name:
>             STRING:
>                 value: Matthias
> ```
> 
> In case of vectors, length, data range and value restriction might be a bit more complicated. But I think this is not a huge priority to be implemented since we are ontologizing in the EMMO-way, without any vectors as data properties.
> 
> Since you already closed this PR and deleted the branch, I can come up with a new branch with this idea.
> 
> What do you think?

Sorry I was not expecting a new message :sweat_smile: 

If you open a new PR and provide a working implementation for SimPhoNy v3 for what you are mentioning, **I will definitely not decline to merge it**. But before you put your hands on the keyboard, I again strongly advise you to carefully check if either [SHACL](https://www.w3.org/TR/shacl/) or [OTTR](https://ottr.xyz/) solve your problem. 

I still did not carefully check, but it is my intention. I prefer to stick to something that exists (as far as I know there are validators for SHACL), especially if it is open-source, rather than to reinvent the wheel. The point I want to make here is that if you implement now a solution for SimPhoNy v3, unless it is better than SHACL or OTTR, I will not keep this solution in v4 or future versions. But unfortunately I do not know if those two can satisfy our needs right now or not.

MBueschelberger:
Hi @kysrpex, thanks for the hint on SHACL and OTTR. 

This functionality for the schema validation was originally requested for a unit test in a small use case, not for a proper pipeline in a larger framework.

On the one hand, I totally agree that SHACL offers good stability+flexiblity against our current validation-algorithm in order to make a proper analysis of our CUDS. 

But on the other hand, I think that a proper integration would require a parser which is able to convert the yml-schema into the [node-shapes of shacl](https://github.com/RDFLib/pySHACL/blob/347a706dce5ea0be8c3c9baf540aa17e3487b3b3/examples/two_file_example.py#L12), which might require some work. 

For me, it is currently not clear in how far this might fit into the timeline of the usecase, since there is enough other work to do (as usual). On the other hand, this is of course a major benefit for SimPhoNy v3 and v4, as you already evaluated. 

I think I have to discuss this internally before deciding which way to go for the time given.

PR #830, opened 2022-11-15
Enh/schema validation attribute values

### LINKED ISSUE #828: Enh: add checking of attribute-values (=values of data properties)
As a follow-up of #821, I was also asked for the functionality to check the actual values of the Cuds.

For my opion this makes absolutely sense if you want to write a unittest for your function which is generating Cuds with actual data.

However, my suggestion would be that this might also overwrite the cardinality with a value-keyword in the case of data properties:

Restriction to length

    city.Neighborhood:
        city.name:
            STRING:
                value: 1+

Following a similar pattern as the cardinality, this value might e.g. in case of a string restrict the length (see example above) or the actual value:

Restriction of value

    city.Neighborhood:
        city.name:
            STRING:
                value: Stühlinger

In the case of floats, bools and ints, there is no restriction of the length needed, but rather a statement towards the data range:

Restriction of range

    city.Citizen:
        city.age:
            INTEGER:
                value: 0+

... or the actual value:

Restriction of value

    city.Citizen:
        city.age:
            INTEGER:
                value: 29
        city.name:
            STRING:
                value: Matthias

In case of vectors, length, data range and value restriction might be a bit more complicated. But I think this is not a huge priority to be implemented since we are ontologizing in the EMMO-way, without any vectors as data properties.

MBueschelberger:
closes #828 

New functionalities for checking the value and length (if it is string) or range (it is a float/int/bool) of a CUDS-attribute.

Please see the updated unittests as reference for this example:

```
city.Citizen:
    city.name:
        STRING:
            length: 1+
            cardinality: 1
    city.age:
        INT:
            range: 0+
            cardinality: 1
```

In the case of a string for an attribute, we can state what should be the length of a `city.name`. For all other types, we can specify the numerical range.

Here is an example of how to restrict a cuds with respect to a certain value, which is very useful for unittesting:

```
city.Citizen:
    city.name:
        STRING:
            value: peter
            cardinality: 1
    city.age:
        INT:
            value: 18
            cardinality: 1
```

Furthermore, the schema validation received an extra keyword-argument called `strict_check` which - as the name already says - triggers a strict check of the cuds during the validation. This basically means that the test fails when any extra cuds apart the ones listed in the schema-file are encountered while querying the root object.

**Please note**: This implementation is closely imitating what is supposed to be covered by any standardized markup-language such as `SHACL`. But due to lack of time in the use case, we could not really focus on integrating these languages into `simphony-osp`, because the effort for translating the schema file into the required shapes was too high.


MBueschelberger:
@kysrpex , there seems to be some problem with the sqlite-wrapper in the pipeline. I am not sure if this is related to the native `sqlite3`-package from python3.

osp/core/utils/schema_validation.py
```diff
@@ -28,6 +29,8 @@ def validate_tree_against_schema(root_obj, schema_file):
         root_obj (Cuds): The root CUDS object of the tree
         schema_file (str): The path to the schema file that
             defines the constraints
+        strict_check (bool): whether extra cuds not listed
```
kysrpex:
```suggestion
        strict_check (bool): whether extra cuds not listed in
```

osp/core/utils/schema_validation.py
```diff
@@ -130,6 +136,65 @@ def _check_cuds_object_cardinality(origin_cuds, dest_oclass, rel, constraints):
...
+                but we found '{}'. The uid of the affected
+                cuds_object is: {}""".format(
+                    str(origin_cuds.oclass),
+                    dest_oclass,
+                    rel_entity,
+                    value,
+                    attribute,
+                    origin_cuds.uid,
+                )
+                raise ConsistencyError(message)
```
kysrpex:
```suggestion
        if value and attribute != value:
            message = """Found invalid attribute value
            between {} and {} with relationship {}.
            The constraint says it should be valued '{}',
            but we found '{}'. The uid of the affected
            cuds_object is: {}""".format(
                str(origin_cuds.oclass),
                dest_oclass,
                rel_entity,
                value,
                attribute,
                origin_cuds.uid,
            )
            raise ConsistencyError(message)
```

kysrpex (changes requested):
Hi @MBueschelberger,

I requested two very minor changes that you can automatically apply just clicking a button on GitHub, if you do it then the PR is good to go.

I am not very happy with the code (and foresee bugs) but this is strongly related to the fact that I am not happy with the original code either. I know that you are just plugging your feature into it and it seems reasonable to me, and the changes do not make things worse. Let's say that I would not be ready to document and promote the whole schema validation feature in its current state (neither was I ready before), but I think this PR is a net positive anyway.

MBueschelberger:
Why does the pipeline fail again after you merged dev into this branch at commit `d3bc34e2b9ae9d5b88c7fa092ba2071f288fc45d`? The pipeline for commit `1a42f0f6731014c43b8ebc2b4e81963f19a8cfe2` was successful but **not** for the following `15a168c3afcde72fd2094e89ce67afc234155aad`? This is confusing

kysrpex (approved):

kysrpex:
> Why does the pipeline fail again after you merged dev into this branch at commit `d3bc34e2b9ae9d5b88c7fa092ba2071f288fc45d`? The pipeline for commit `1a42f0f6731014c43b8ebc2b4e81963f19a8cfe2` was successful but **not** for the following `15a168c3afcde72fd2094e89ce67afc234155aad`? This is confusing

Good question honestly, I will try to find out.

